PR #96, opened 2015-01-18
varpart plot: new 4-part plot and colour scheme

jarioksa:
This PR re-designs 4-part plots of `varpart` as four ellipses instead of three circles and two rectangles. The new design is picked up from Legendre, Borcard & Roberts, *Ecology* **93**,1234-1240 (2012), and looks like this in `example(varpart)`:
varpart4

The PR also adds arguments for background colour and its transparency.

I also considered adding an option of labelling the component circels/ellipses, but my attempts had too many lines of code and too many conditions (`if`)  to make a neat function, and therefore this was skipped.

### DIFF
--- a/R/showvarparts.R
+++ b/R/showvarparts.R
@@ -9,12 +9,15 @@
             bg <- rep(bg, length.out = parts)
     }
     cp <- switch(parts,
-                 c(0,0),
-                 c(0,0, 1,0),
-                 c(0,0, 1,0, 0.5, -sqrt(3/4)),
-                 c(NA,NA)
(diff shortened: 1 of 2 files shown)


jarioksa:
Pierre Legendre privately sent revised functions that add configurable names to the circles/ellipses. Most things seem to be OK now, except that `showvarparts`  argument `cex.main`  fails when `showvarparts`  is called from `plot.varpart`: argument `cex` cannot be used to set the sizes of fraction labels but it is expanded to `cex.main`  that changes the source names. Either `cex.main`  must be renamed or `cex`  added to the `show.varpart`  code.

Issue #100, opened 2015-02-23
Scoping issue in `anova.ccabymargin`


gavinsimpson:
Not sure *if* we can do much about this now or in the future, but I'm recording it here in case we update something in the future?

The following example illustrates a scoping bug/infelicity that, ideally, we'd be able to handle in **vegan**

```r
library("vegan")
data(dune, dune.env)
foo <- function(x, env) {
    m <- rda(x ~ Manure + A1, data = env)
    anova(m, by = "margin")
}
out <- lapply(dune, foo, env = dune.env)
```

which fails with:

```r
> out <- lapply(dune, foo, env = dune.env)
Error in eval(expr, envir, enclos) : object 'x' not found
```

the `traceback()` is:

```r
> traceback()
18: eval(expr, envir, enclos)
17: eval(expr, p)
16: eval.parent(specdata, n = envdepth)
15: ordiParseFormula(formula, data, na.action = na.action, subset = substitute(subset))
14: rda.formula(formula = x ~ A1, data = env)
13: rda(formula = x ~ A1, data = env)
12: eval(expr, envir, enclos)
11: eval(call, parent.frame())
10: update.default(object, paste(".~.-", nm))
9: update(object, paste(".~.-", nm))
8: permutest(update(object, paste(".~.-", nm)), permutations, ...)
7: FUN(c("Manure", "A1")[[1L]], ...)
6: lapply(trmlab, function(nm, ...) permutest(update(object, paste(".~.-", 
       nm)), permutations, ...), ...)
5: anova.ccabymargin(object, permutations = permutations, model = model, 
       parallel = parallel, scope = scope)
4: anova.cca(m, by = "margin")
3: anova(m, by = "margin") at #3
2: FUN(X[[1L]], ...)
1: lapply(dune, foo, env = dune.env)
```

The obvious workaround is

```r
outs <- vector(mode = "list", length = NCOL(dune))
for (i in seq_len(NCOL(dune))) {
    resp <- dune[, i, drop = FALSE]
    m <- rda(resp ~ Manure + A1, data = dune.env)
    outs[[i]] <- anova(m, by = "margin")
}
```

or, retaining `lapply()`:

```r
bar <- function(i, env) {
    m <- rda(dune[,i] ~ Manure + A1, data = env)
    anova(m, by = "margin")
}
out <- lapply(seq_len(NCOL(dune)), bar, env = dune.env)
```

jarioksa:
[...]

PR #101, opened 2015-02-24
Scoping issues in anova.ccaby* methods

jarioksa:
There were too almost simultaneous reports on scoping issues. Benedicte Bachelot reported that `anova.ccabyterm` failed when the community data had a name of a function. She had `all`, and the test case has
```R
data(dune, dune.env)
cca <- dune
m <- cca(cca ~ ., dune.env)
anova(m, by="margin") # fails
```
A couple of days later, Richard Telford tweeted about problems in embedding `anova.ccabymargin` in `lapply`, and @gavinsimpson raised this as issue #100. When testing this, I noticed that the fix made for the Bachelot case also corrected this issue. 

I am not quite sure that this PR is completely safe: environments are a pain. However, all old test cases pass and some that failed earlier also work now. Some scoping issues still remain, but the reported bad cases were corrected. Better make this public for review, though.

### DIFF
--- a/R/capscale.R
+++ b/R/capscale.R
@@ -16,8 +16,8 @@
     ## The following line was eval'ed in environment(formula), but
     ## that made update() fail. Rethink the line if capscale() fails
     ## mysteriously at this point.
-    X <- eval(formula[[2]], envir=parent.frame(),
-              enclos = environment(formula))
+    X <- eval(formula[[2]], envir=environment(formula),
+              enclos = globalenv())
     if (!inherits(X, "dist")) {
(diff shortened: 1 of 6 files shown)


gavinsimpson:
@jarioksa This looks good and the fix (evaluating in the environment of the formula) is mentioned as a fix to the "bug" in `update()`, which evaluates in the `parent.frame()`, that Hadley mentions in his [Advanced R book section on capturing calls](http://adv-r.had.co.nz/Expressions.html#capturing-call).

I'm going to merge this...

PR #105, opened 2015-03-06
Fix goodness functions for partial constrained ordination 

jarioksa:
This PR fixes problems emerging when inspecting an [R-sig-ecology message](https://stat.ethz.ch/pipermail/r-sig-ecology/2015-March/004936.html) by Christoph Eberhard:

1. The variance of partial component was overestimated by a factor of *N-1* in `goodness.rda`. 
2. `goodness.rda`  and `goodness.cca`  added the partial component of the variation both to the constrained (`model = "CCA"`) and residual (`model = "CA"`) component so that it was added twice. For any constrained ordination model `m`, the following should give a vector of ones: `goodness(m, model = "CCA", sum=TRUE) + goodness(m, mod = "CA", sum=TRUE)`, but that failed in partial models.
3. If only one axis was chosen, the result dropped to a vector and `goodness.cca`  and `goodness.rda` gave warnings.
4. The results were inconsistent when only some axes were chosen. Now cumulative sums are evaluated prior to choosing axes so that the proportion explained by axis 'k' no longer depends on the axes selected before 'k'.

It seems that all these problems were present when these functions were released in **vegan** 1.6-7 on Jan 24, 2005 so that they have persisted longer than a decade.

### DIFF
--- a/R/goodness.cca.R
+++ b/R/goodness.cca.R
@@ -39,7 +39,8 @@
                 Xbar <- t(Xbar)
             ptot <- diag(crossprod(Xbar))
             tot <- tot + ptot
-            vexp <- sweep(vexp, 1, ptot, "+")
+            if (model == "CCA")
+                vexp <- sweep(vexp, 1, ptot, "+")
         }
         vexp <- sweep(vexp, 1, tot, "/")
     }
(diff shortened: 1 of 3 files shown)


jarioksa:
Was OK in further testing. It seems that `goodness` refuses to calculate `statistic="distance"` with `display="sites"` for constrained ordination, but this also could be done -- we do it in `stressplot`.

Issue #110, opened 2015-05-06
pcnm should fail earlier if passed something clearly not a distance matrix


gavinsimpson:
This is a little unfriendly:

```r
> data(mite.xy)
> foo <- pcnm(mite.xy)
Error in as.matrix.dist(d^2) : 
  length of 'dimnames' [1] not equal to array extent
```

Suggest we change `pcnm` to have the following check

```r
if (!inherits(dis, "dist")) {
    dims <- dim(dis)
    if (length(unique(dims)) >1) {
        stop("'dis' doesn't appear to be a square distance matrix.")
    }
    dis <- as.dist(dis)
}
```

Thoughts?

jarioksa:
Closed with PR #113

PR #113, opened 2015-05-06
Fail earlier if 'dis' is not class 'dist' or a square matrix-like object

gavinsimpson:
This addresses the issue raised in #110 and implements a check for a square matrix if `dis` does not inherit from class `"dist"`. The check now throws an error indicating a non-square matrix if this check fails.

This avoids the rather cryptic error message that is thrown much later in the current code after the `wcdmscale()` call.

### DIFF
--- a/R/pcnm.R
+++ b/R/pcnm.R
@@ -1,6 +1,11 @@
-`pcnm` <-
-    function(dis, threshold, w, dist.ret = FALSE)
-{
+`pcnm` <- function(dis, threshold, w, dist.ret = FALSE) {
+    if (!inherits(dis, "dist")) {
+        dims <- dim(dis)
+        if (length(unique(dims)) >1) {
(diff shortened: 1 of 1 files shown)


jarioksa:
This looks OK. It is a bit funny to see how many ways we have implemented to check if input could be interpreted as dissimilarities. In `bioenv.default` (snippet below) and `metaMDS` we also check that the matrix is symmetric:
```R
if (is.matrix(comm) && nrow(comm) == ncol(comm) & isTRUE(all.equal(comm, t(comm)))) {
        comdis <- as.dist(comm)
       index <- "supplied square matrix"
 }
```
In `isomapdist` we check that the input can be transformed to a `"dist"` object without warnings:
```R
op <- options(warn = 2)
on.exit(options(op))
if (!inherits(dist, "dist"))
    dist <- as.dist(dist)
options(op)
```
Most often we just ignore any tests and just run `as.dist()`.

Issue #102, opened 2015-02-25
Export `ordiArrowMul()`


gavinsimpson:
Would there be any objection to exporting the currently internal function `ordiArrowMul()`? **ggvegan** makes use of this to rescale arrows in (tri|bi)plots of constrained ordinations. I'm currently accessing this as `vegan:::ordiArrowMul` but `R CMD check` complains about this usage as a Note.

If it's not suitable for export in its current form, what would be required before we'd be happy exporting it?

jarioksa:
Gav,

It is quite OK to export ordiArrowMul. It was not exported because I thought it was not so interesting to the users, but actually we tell how to use it in docs so that we also could export it.

NB, we use now ordiArrowTextXY to calculate the coordinates of text labels at the tip of the arrow. Should this be exported, too?

Cheers, Jari

gavinsimpson:
Great. I'll work on that, and I do think it would be useful to export `ordiArrowTextXY` - I wonder what I am doing to add text labels to arrows in **ggvegan**...?

I'll self-assign this to me and work through the details.

jarioksa:
I merged PR #107. However, there are a couple of things to do after this:

I moved the `ordiArrow*` documentations from `man/ordiarrows.Rd` to `man/envfit.Rd` in commit 24776d6b4. Despite similarity of names, `ordiarrows()` function is unrelated to `ordiArrow*` functions. The `ordiArrow*` functions are only used in `plot.envfit()` and `plot.cca` and their friends, and the functions should be documented with these. However, I think that the current example may be a bit long and technical in the context of `envfit`  documentation. 

The `ordiArrow*`  functions were internal functions, and therefore they are not user-friendly: we assume that calling functions take care that calls to internal functions are absolutely correct. If we let loose these functions, we should add some elements of usability. I think the minimal thing is that the input data `x`  can also be an ordination object or an `envfit` object, and `ordiArrowMul`  can extract the scores and `ordiArrowTextXY` can extract the scores and labels from this object. This could happen by adding `x <- scores(x, dis="sites", ...)` to these functions.

jarioksa:
It perhaps makes sense to separate `ordiArrow*` documentation into its own `Rd` page. This would allow nerdier description and examples but reduce clutter in `man/envfit.Rd`. I think `envfit()` is an entry-level function which should be explained in simple terms, and technical details should be explained somewhere else. 

Just my €0.02.

gavinsimpson:
[...]

PR #114, opened 2015-05-06
Make ordiArrowMul and ordiArrowTextXY more user-friendly

gavinsimpson:
This PR and set of commits addresses comments/suggestions in #102 to make `ordiArrowMul()` and `ordiArrowTextXY()` more user friendly in terms of the objects they work with and simplifying extraction of scores and labels without the user having to do this ahead of time.

This PR also pulls the documentation for these functions out into their own `.Rd` file: `ordiArrowTextXY.Rd` and removes the relevant sections from `envfit.Rd`.

Points to note:

 1. If `x` inherits from class `"envfit"` then I hard code what type of scores to extract: `display = "vectors"`. This seemed reasonable as that is all I would expect you'd want to plot with these functions
 2. I chose not to supply a default for `display` in all other cases as this should really be specified by the user; I couldn't think of a good argument as to why we'd want to anticipate this.

Thoughts?

### DIFF
--- a/R/ordiArrowMul.R
+++ b/R/ordiArrowMul.R
@@ -1,19 +1,33 @@
 ### Scaling of arrows to 'fill' a plot with vectors centred at 'at'.
 ### Plot dims from 'par("usr")' and arrow heads are in 'x'.
-`ordiArrowMul` <-
-    function (x, at = c(0,0), fill=0.75) 
-{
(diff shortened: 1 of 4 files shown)


jarioksa:
Looks good, and seems not break things.

gavinsimpson:
Thanks @jarioksa (I didn't just want to merge this until you;d looked at it and were happy with decisions I'd made.)

PR #115, opened 2015-05-17
Rarefy slope

jarioksa:
This started as a simple enhancement to fulfill the wish to find the slope of rarefaction curve at given sample size (https://stat.ethz.ch/pipermail/r-sig-ecology/2015-May/005038.html). This radiated into a bigger set of changes that now includes:

* `rareslope`: the function originally suggested for the slope of `rarecurve` (or the derivative of `rarefy`). 
* `specslope`: similar functions for species accumulation curves (`specaccum`) when these are based on analytic models and allow finding derivatives (methods `"exact"`, "`rarefaction"` and `"coleman"`). 
* `specslope` was made generic, and also finds the derivatives of `nls` regression models implemented in `fitspecaccum`. 
* `predict.specaccum`  model uses now analytic equations for `"exact"`, "`rarefaction"` and `"coleman"` methods). These allow using non-integer `newdata`.  Formerly we only interpolated from fitted values, but this is now done only when analytic equations are unavailable.
* `fitspecaccum` models gained `AIC` and `deviance` methods for their fitted `nls` models.


jarioksa:
I found new issues that need fixing: handling of weights 'w' is haphazard.

PR #118, opened 2015-05-28
One point hulls and ellipses

jarioksa:
This PR gives a simple solution to a problem raised in StackOverflow http://stackoverflow.com/questions/30487677/ordihull-label-with-single-occurance about labelling one-point convex-hulls in `ordihull`. With this PR the hulls are labelled but nor drawn nor returned. The same enhancement was also made to `ordiellipse`: one-point ellipses are labelled, but not drawn nor returned. It could be argued that the (invisible) return object has information on these one-point cases as well, but that requires a bit more work, because `summary` and `ordiareatest` should work for these one-point cases, too. 

Function `ordispider` already works ideally: it labels one-point spiders but does not draw them. However, it also returns the "centroid" of one point (or the coordinates of that one point).

### DIFF
--- a/R/ordiellipse.R
+++ b/R/ordiellipse.R
@@ -37,8 +37,15 @@
     kk <- complete.cases(pts) & !is.na(groups)
     for (is in inds) {
         gr <- out[groups == is & kk]
-        if (length(gr) > 1) {
-            X <- pts[gr, ]
+        if (length(gr)) {
+            X <- pts[gr, , drop = FALSE]
+            ## one-point ellipses are labelled but not returned in the
+            ## result object (nor calculated)
+            if (length(gr) == 1) {
+                cntrs <- rbind(cntrs, X)
+                names <- c(names, is)
+                next
+            }
             W <- w[gr]
             mat <- cov.wt(X, W)
             if (kind == "se")
--- a/R/ordiellipse.R
+++ b/R/ordiellipse.R
@@ -39,21 +39,19 @@
         gr <- out[groups == is & kk]
         if (length(gr)) {
             X <- pts[gr, , drop = FALSE]
-            ## one-point ellipses are labelled but not returned in the
-            ## result object (nor calculated)
-            if (length(gr) == 1) {
-                cntrs <- rbind(cntrs, X)
-                names <- c(names, is)
-                next
-            }
             W <- w[gr]
             mat <- cov.wt(X, W)
+            if (mat$n.obs == 1)
+                mat$cov[] <- 0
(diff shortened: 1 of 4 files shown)


gavinsimpson:
We could be more efficient in building up `cntrs` and `names` than `rbind`ing them. This [line](https://github.com/vegandevs/vegan/pull/118/files#diff-c69339891911635e52ac2caab2a49a37R34) seems redundant:

```r
if (label)
         cntrs <- names <- NULL
```

as we could just create them anyway. Won't `label = FALSE` cause a failure later as `cntrs` doesn't exist [here](https://github.com/vegandevs/vegan/pull/118/files#diff-c69339891911635e52ac2caab2a49a37R45):

```r
cntrs <- rbind(cntrs, X)
names <- c(names, is)
next
```

We should allocate `cntrs` and `names` to the required length - we will always fill in a non-zero length group now - and then do the proper thing in the loop to assign into elements rather than grow an object. Even if the efficiency is unlikely to matter much here given there will be few growth increments

jarioksa:
You are correct: `if()` needs `label &&`.

I wouldn't care about efficiency of `rbind` and `c`. It's all about PC and Do The Right Thing(TM). You can't measure the difference.

gavinsimpson:
Exactly; I'd be more comfortable (given my repeated chastisements of users on StackOverflow) if we did the right thing and not adopt a *do as I say not as I do* approach :-)

jarioksa:
Yep. The things can be done more cleanly than we do now. Both `ordihull` and `ordielipse` **almost** handle one-point groups. In `ordihull` we need to take care that the internal (to `ordihull`)  `polycentre` does not drop dimensions, and in `ordiellipse` the `cov.wt` gives `NaN` as covariances of one row, and we should make this `0` instead, and `vegan::::veganCovEllipse` should be made to handle such cases. Then these would draw points (lines or polygons with all equal coordinates).

jarioksa:
I commited cleaner fixes. Now we treat one-point hulls and ellipses like any other group. In `ordihull` we only need take care that one-point data do no drop dimensions, and in `ordiellipse`  that the covariance matrices have all zero elements.

The code could be cleaned now like you outlined in your response (pre-allocate `cntrs`  and `names` instead of growing them), and I think we no longer need test of group size (`if(length(gr))`: or could `gr` be of zero-length?). I don't have an opportunity to fix these before next week. @gavinsimpson : if you feel these fixes are OK and you want to fix those things, you can merge this PR and go on with the changes. In other cases I'll have a look at this someday next week.

PR #121, opened 2015-06-24
Consensus RDA

guiblanchet:
Added the function I wrote for the ordiconsensus package.

The RV(), coeffcompare() and consensusRDA() functions are the ones that are necessary to perform a consensus RDA as is presented in Blanchet et al. (2014) in Ecological Monograph. 

I also added a function called simulSADcomm() which was used to simulate data for the paper. This function does not need to be included in vegan, but in case someone feels it is appropriate to be there... well.. it is ! 

Finally, I also added a function called SADbin() which, I think, is comparable to prestonfit() and fisherfit() already in vegan. In essence, SADbin() can be used to construct species abundance distributions innings (or octaves). The main difference, is that SADbin() is very general. Again, this function does not need to be included to perform a consensus RDA, but the code in this function may be used generalize prestonfit() and/or fisherfit()


jarioksa:
I cannot check these changes now: battery charger of my laptop broke down and the power ran out. I write these notes on iPad which cannot be used for R development. I had some notes on these functions in that laptop, but I cannot access them now. The changes suggested here are isolated: they do not change other vegan functions, and could be safely merged. I have reservations for the SAD and simulation functions which probably should be better integrated with other vegan functions. For that, I would like to see my notes in my laptop.

jarioksa:
@guiblanchet first comments on `consensusRDA`:

From the **vegan** tradition, the function puts too much responsibility to the user. The function needs a list of `"rda"` models, a matching list of `anova.cca(…, by = "axis")` models and a matrix of constraints, but it does not check that these match -- `anova` models can be in different order than the `rda` models, and the function accepts `Y` data that is different than in `rda`. It seems that right hand sides of `rda` models should be identical, but this is not tested. People do misuse functions, and basic misuse should be detected. One peculiar thing is that `rda` and `capscale` models accept factor constraints but `consensusRDA` rejects them. 

Many of these things can be handled internally:

1. We have `model.matrix.cca` function that can handle `rda` and `capscale` results. For factor constraints this returns their numerical contrasts and allows analysing non-numerical constraint matrices.

2. With `model.matrix.cca` we can also drop argument `Y`.

3. Further, we can check that model formulae have identical RHS.

4. I think that arg `ordisigniaxis` could be either a numerical vector, or if missing, the `anova` could be evaluated within the function so that the tests are really made to the correct models in the correct order. So we could have something like `lapply(ordires, anova, by = "axis", cutoff = pval, …)`. Here the last dots `…` allows passing `parallel` argument to the tests (and then we need to add dots to the `consensusRDA` call). 

5. No **vegan** function should use matrix `det()`. It is never needed but can always be replaced with sound numerical coding. In addition to being numerically poor, the current code is wrong and does not find the correct rank. What you do is that you remove the first variable (column) as long as the determinant is numerically zero (below the threshold). Assume you have a non-singular matrix `X` of, say, rank=7. If you make a matrix where one column is duplicated, say, `X[,c(1:4,4:7)]`, this will remove columns before the duplicated column, and give rank=4 matrix. The canonical way of finding the rank is to use QR decomposition as `qr(Y)$rank`. If you think you also need a full-rank matrix, you can use
```R
Q <- qr(Y)
Y <- Y[, Q$pivot[seq_len(Q$rank)], drop = FALSE]
```

I think it would make sense to return an object of class `"consesusRDA"` instead of simple list so that we could provide method functions. I also think that such method functions should be provided - we usually do so in **vegan**. At least `print` and `plot` are useful (and probably `summary` as well).

guiblanchet:
I was not aware that you wrote the ```model.matrix.cca``` function and this seems like it is definitely worth using. I will work on making the different modifications you proposed to make them **vegan** ready. Below are some more direct answers to your 5 comments:

1 - Perfect ! I will try to see how far I can go with this.

2 - Perfect !

3 - After reading your comments and playing a little bit with  the ```model.matrix.cca``` function, this was my idea.

4 - When I originally build that function, my idea was that ```ordisigniaxis``` could *either* be a numerical vector *or* an ```anova``` object, not both so that the test of canonical axes for each RDAs would only need to be ran once. It would also prevent for saving the list of ```anova.cca``` objects. To follow the **vegan** tradition, would it be better to change this feature so that a list of ```anova.cca``` object would be required ?

5 - I fully agree with this ! I will make the change

Schedule wise, I will try to get back to you with the new changes by late next week.

jarioksa:
@guiblanchet - about point 4: it makes sense to accept a list of `anova.cca` models as input to `ordisigniaxis`. However, it should be made very clear that only `by = "axes"` can be used (and perhaps this should be tested). Currently `anova.cca` functions return a `data.frame` of results with information on header text, but there is no easy and robust way to detect which `by` case (if any) was used. Detecting this by the name of the variable (like you do now) is non-robust, and relying on header text is fragile. Possibly we should add a `by` argument to the `anova.cca` results.

Issue #122, opened 2015-06-30
permustats for anova.cca


EDiLD:
I just noticed that there is no permustats method for anova.cca.

Shouldn't be hard to implement:
Just return F.perm from permutest with the anova.cca-object and write a permustats.anova.cca method.

Why is this? Maybe not the clutter the anova.cca object?

gavinsimpson:
Will have to think on this some; what should we do with with `by = "axis"` and `by = "margin"`? Where we would have multiple rounds of sampling distributions depending on how many axes or variables are under test.

A matrix representation would possibly be suitable, with a column per variable/axis?

jarioksa:
We even document that `permustats` is not available for `anova.cca` results but only for `permutest.cca`.  It is possible to change the functions so that they return permutation results for `permustats`, but requires quite a lot of work. If I remember correctly, laziness was the reason not changing these functions.

The `anova.cca` function returns an object of class `anova` and it is displayed with `stats:::print.anova`. The output should be compatible with that function and return only a single ANOVA table. The permutation results could be returned as an attribute that is ignored by `stats:::print.anova`.  

Quite a few functions need changing. We have the following call trees for `permustats.cca` in current **vegan**:
```
anova.cca -> permutest.cca
anova.cca -> anova.ccalist -> permutest.cca
anova.cca -> anova.ccabyterm -> anova.ccalist -> permutest.cca
anova.cca -> anova.ccabymargin -> permutest.cca
anova.cca -> anova.ccabyaxis -> permutest.cca
```
All these cases should be covered. This is possible but requires work. This is not among my top priorities, but pull requests are welcome (after critical review).

EDiLD:
> A matrix representation would possibly be suitable, with a column per variable/axis?

@gavinsimpson Jupp, that is what is done e.g. in `adonis()`.

@jarioksa I know that it's documented. I'm just using `anova.cca` more than `permutest` and though it would be a nice plus...

> All these cases should be covered. This is possible but requires work. This is not among my top priorities, but pull requests are welcome (after critical review).

I'll have a longer train trip this weekend (to [SEFS 9](http://www.sefs9.ch/)) -> I'll try to implement this.

BTW:
We will give a short workshop @SEFS ("*Data analysis in freshwater ecology using R*"), where I will present ordinations using the vegan package. You can find my materials [here](https://github.com/EDiLD/sefs9_Rworkshop/tree/master/2-Ordination).

jarioksa:
[...]

PR #123, opened 2015-07-02
permustats.anova.cca

EDiLD:
This PR implements permustats.anova.cca (Issue #122).


EDiLD:
Thanks.
Revised according your comments.

PR #131, opened 2015-08-17
New adonis

jarioksa:
This is a new alternative implementation of `adonis` based on `anova.cca`. The main advantage is that this PR will provide marginal and overall tests in addition to the sequential tests of old `adonis`. The main disadvantage is that this runs much more slowly. Therefore I provide the new test as an alternative function `adonis2`  and keep the old `adonis`.

The new `adonis2`  is the public and exported function. Basically, it contains little more than an internal function `adonis0`  that creates a minimal `"capscale"` like object and `anova`  call that runs the test with that minimal `"capscale"` object. The `adonis0`  result contains only items that are necessary to run sequential or marginal `anova`, but it skips all eigenanalysis and is therefore faster than full-blown `capscale`. However, since `adonis0` or `capscale` are run only once per permutation cycle, the time saving is rather small (about 5% in my tests), and most time is spent in `anova.cca`. It seems that the sequential tests in old `adonis`  are very efficient. In `anova.cca`  we need to run separate permutations for each term and this is much slower. However, `anova.cca` benefits much from parallel processing, whereas `adonis`  seems to suffer from parallel processing. 

The real question is: should this be merged? We do get marginal and overall tests, but with sequential tests of old `adonis` there is no benefit.

### DIFF
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -38,6 +38,8 @@ export(anova.cca)
 export(as.mcmc.oecosimu, as.mcmc.permat)
 ## export oldCapscale for compatibility after new capscale design
 export(oldCapscale)
+## alternative implementation of adonis: may be eliminated later
+export(adonis2)
 
 ## export regular functions with dot names
 
--- a/R/adonis2.R
+++ b/R/adonis2.R
@@ -1,6 +1,6 @@
 `adonis2` <-
-    function(formula, data, method = "bray", by = "term",
-             permutations = 999, parallel = getOption("mc.cores"), ...)
+    function(formula, data, permutations = 999, method = "bray",
+             by = "term", parallel = getOption("mc.cores"), ...)
 {
     environment(formula) <- environment()
     sol <- adonis0(formula, data = data, method = "bray")
--- a/R/adonis2.R
+++ b/R/adonis2.R
@@ -4,8 +4,13 @@
 {
     environment(formula) <- environment()
     sol <- adonis0(formula, data = data, method = "bray")
-    anova(sol, permutations = permutations, by = by,
-          parallel = parallel)
+    out <- anova(sol, permutations = permutations, by = by,
+                 parallel = parallel)
+    ## Fix method name in output
+    head <- attr(out, "heading")
+    head[2] <- sub("adonis0", "adonis2", head[2])
+    attr(out, "heading") <- head
+    out
 }
 `adonis0` <-
     function(formula, data=NULL, method="bray", ...)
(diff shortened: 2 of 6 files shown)


EDiLD:
>The real question is: should this be merged? We do get marginal and overall tests, but with sequential tests of old adonis there is no benefit.

For me: The gains (+marginal +overall tests) out-weight the losses (-speed).
Many (most?) users have small datasets, so the losses aren't very big.

On my (lame) laptop:
```
system.time(
+   ad <- adonis(dune ~ Management*A1, data=dune.env, permutations=999)
+ )
   user  system elapsed 
  0.183   0.320   0.150 
d> 
d> system.time(
+   ad2 <- adonis2(dune ~ Management*A1, data=dune.env, permutations=999)
+ )
   user  system elapsed 
  0.620   0.425   0.577 
```

So 4x slower, but in absolute values 0.4 seconds slower. 
I don't care about these 0.4 seconds - I even can't take a sip of coffee in this time...

Same for the bigger brycedata:
```
d> system.time(
+   ad <- adonis(bryceveg ~ pos*quad, data = brycesite, permutations = 999)
+ )
   user  system elapsed 
  2.489   0.322   2.480 
d> 
d> system.time(
+   ad2 <- adonis2(bryceveg ~ pos*quad, data = brycesite, permutations = 999)
+ )
   user  system elapsed 
  7.616   1.071   7.533 
```

5 seconds difference, but for me still acceptable.

You could keep the old adonis as faster alternative (if you really need the speed), but then I would also change the names, making adonis2 as default (`'adonis2'` -> `'adonis'` and `'adonis'` -> `'.adonis'`? [the dot notion is similar to base-R (eg. lm.fit, colSums, ...)]).

jarioksa:
My relative timings are similar: for small data sets (`mite`, `dune`), new `adonis2` takes 3 to 4 times longer time to run. Parallel processing with four processors reduced the slow-down to <2 times. Curiously, old `adonis` sufferred slightly from parallel processing, and `adonis2` more than doubled its speed. 

Almost all time was used in `anova.cca`. One profiling run was barely able to detect `adonis0` which used <2% of total time. Speed-up there does not help. It could help to write  `permutest.cca` as compiled code, but I think this would still be the slow part. In my profiling, >50% of time was spent in the FORTRAN code of `qr.fitted` which already is compiled code. So it is the other <50% that could benefit from new coding. 

Compiling R into binary code with **compiler** package has little effect. Speed-up is <5% in all cases. This is my general experience: JIT compiler does not help much.

Issue #133, opened 2015-08-24
metaMDS() fails when computing wascores if supplied a long expression for `comm`


gavinsimpson:
Here's a reproducible example:

```r
library("vegan")
data(varespec)
set.seed(1)
metaMDS(subset(varespec, select = colSums(varespec) > 0, subset = rowSums(varespec) > 0))
```

which produces

```r
> metaMDS(subset(varespec, select = colSums(varespec) > 0, subset = rowSums(varespec) > 0))
Square root transformation
Wisconsin double standardization
Run 0 stress 0.1843196 
Run 1 stress 0.2455912 
Run 2 stress 0.2169407 
Run 3 stress 0.2313231 
Run 4 stress 0.1974406 
Run 5 stress 0.1858402 
Run 6 stress 0.1948414 
Run 7 stress 0.2265717 
Run 8 stress 0.222507 
Run 9 stress 0.2023215 
Run 10 stress 0.2673177 
Run 11 stress 0.1976151 
Run 12 stress 0.1852397 
Run 13 stress 0.2341085 
Run 14 stress 0.1955867 
Run 15 stress 0.2137409 
Run 16 stress 0.2109638 
Run 17 stress 0.1825658 
... New best solution
... procrustes: rmse 0.04169825  max resid 0.1521436 
Run 18 stress 0.1843197 
Run 19 stress 0.2570119 
Run 20 stress 0.3760596 
Error in parse(text = attr(dis, "commname")) : 
  <text>:1:94: unexpected ')'
1: wisconsin(sqrt(subset(varespec, select = colSums(varespec) > 0, subset = rowSums(varespec) > )
```

The problem occurs [here](https://github.com/vegandevs/vegan/blob/master/R/metaMDS.R#L50) as we try to recover the possibly transformed data from the calling frame:

```r
    if (wascores) {
        ## transformed data
        comm <- eval.parent(parse(text=attr(dis, "commname")))
        wa <- wascores(points, comm, expand = expand)
    }
```

The failure is due to the `deparse(substitute(comm))` step early on in `metaMDS()` breaking the expression up into more than a single string:

```r
foo <- function(comm) {
    deparse(substitute(comm))
}
foo(subset(varespec, select = colSums(varespec) > 0, subset = rowSums(varespec) > 0))
```

which produces:

```r
> foo(subset(varespec, select = colSums(varespec) > 0, subset = rowSums(varespec) > 0))
[...]

PR #135, opened 2015-08-25
Return & use (possibly modified) community data when computing WA scores

gavinsimpson:
This PR implements a fix for #133. It makes three changes

 1. the (possibly modified) `comm` data is returned from `metaMDSdist()` as attribute `comm` of object `dis`,
 2. computation of WA scores in `metaMDS()` now uses `attr(dis, "comm")` instead of trying to recreate the original data
 3. the deparsed expression passed to the `comm` argument of `metaMDS()` is now robust to very long expressions

### DIFF
--- a/R/metaMDS.R
+++ b/R/metaMDS.R
@@ -6,7 +6,13 @@
               plot = FALSE, previous.best,  ...) 
 {
     engine <- match.arg(engine)
-    commname <- deparse(substitute(comm))
+    ## This could be a character vector of length > 1L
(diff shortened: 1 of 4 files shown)


jarioksa:
Looks sound: merging? 

I guess this should go to 2.3-1? It was reported as a bug.

gavinsimpson:
Yes, it probably should go there. Remind me what the procedure is for that? Do we have to cherry-pick from this commit to apply to the 2.3-x branch?

Issue #132, opened 2015-08-24
permutest.cca breaks tied values (a case of R FAQ 7.31)


jarioksa:
Sven Neulinger reported a problem with `anova.cca`  with tied values. Here an excerpt of his email:

"Since anova.cca() uses permutation tests, there is a lower limit for the
P-value with small numbers of replicates per treatment group (n) even if
the F-value is very large [e.g., Clarke, K.R. 1993. Non-parametric
multivariate analysis of changes in community structure. Australian
Journal of Ecology 18: 117-143]. Specifically, with n=4 and 2 treatment
levels, the minimal P-value produced by a permutation-based ANOVA is
expected to be ~0.02857.

This outcome should be independent of the number of species (OTUs) in
the count table provided to anova.cca(). However, I noticed that
anova.cca() sometimes gives P-values far lower than what could be
expected."

The following code reproduces the problem:
```R
A <- gl(2, 4)
set.seed(2)
y <- c(rnorm(4, mean=0), rnorm(4, mean=10))/7
mod <- rda(y ~ A)
# permutation matrix
p <- shuffleSet(8, control=how(complete = TRUE, nperm=40320, maxperm=50000))
anova(mod, permutations = p)
##Permutation test for rda under reduced model
##Permutation: free
##Number of permutations: 40319
##
##Model: rda(formula = y ~ A)
##       Df Variance      F   Pr(>F)    
##Model     1  0.60590 242.77 2.48e-05 ***
##Residual  6  0.01497                    
```
The reported *P*-value is the  lowest possible value 1/(40319+1) meaning that **all** permutation *F*-values are lower than the observed statistic:
```R
> out <- permutest(mod, permutations = p)
> max(out$F.perm) - out$F.0
        Model 
-9.947598e-13 
> sum(out$F.perm >= out$F.0)
[1] 0
```

We round to 12 decimal places in `permutest.cca` to avoid false breaking of ties, but obviously we need more rounding -- but how much? We do not want to falsely equalize different values. I think we should also use `signif` (giving the requested number of significant digits) instead of `round` (giving the requested number of decimals).

The fix should be simple, but I made this a public issue to raise awareness: `permutest.cca` is not the only function that is similarly affected.

jarioksa:
Closed with PR #136

Issue #120, opened 2015-06-12
Bug in p-value calculations in adonis


fra-gh:
The p-value calculated by adonis is sometimes wrong. 
For example, the following code 
```R
library("vegan")
  options(digits = 12)
  g = as.factor(c("Control","Control","Group1", "Group2", "Group2"))
  d = as.dist(matrix(c(0,3,7,2,1, 3,0,5,4,1, 7,5,0,2,6, 4,2,4,0,2, 1,1,6,2,0), ncol=5))
  pp = adonis(d ~ g)
  pp
```
returns
```

Call:
adonis(formula = d ~ g) 

Permutation: free
Number of permutations: 120

Terms added sequentially (first to last)

          Df SumsOfSqs MeanSqs     F.Model           R2  Pr(>F)
g          2      23.3   11.65 3.584615385 0.7818791946 0.20833
Residuals  2       6.5    3.25             0.2181208054        
Total      4      29.8                     1.0000000000  
```

The p-value is wrong (only 24 permutations with high f  instead of 39). The correct p-value can be computed from the definition to be
```R
(rowSums(t(pp$f.perms >= pp$aov.tab$F.Model[1]))+1) / (length(pp$f.perms) + 1)
```
which correctly returns
```
[1] 0.333333333333
```

It looks like the bug is due to rounding. When comparing f.perms and F.mod, the value of f.perms is rounded but the value of F.mod is not (although it is rounded afterwards). This makes the >= comparison fail in this example because the value of f.perms is rounded down.

jarioksa:
Confirmed: this is a bug. Fixed in commit f727b589e.

Thanks for reporting this.

fra-gh:
thanks for the very prompt reply. keep up the good work!

jarioksa:
PR #136 has a better fix for this issue.

chitrabaniya:
Do we always need to calculate p-value separately or it goes automatically fixed?
I got surprised with different p-values.

PR #136, opened 2015-08-26
Use epsilon with P-value comparisons in permutation tests

jarioksa:
**vegan** has been pretty careless with `>=`  and `<=` comparisons in assessing *P*-values from permutations. Problems arise when simulated values are tied with observed statistics, but floating point numbers can differ slightly. This is the most frequently asked **R** question and **R** FAQ 7.31. Some functions (`permutest.cca`, `permutest.betadisper`) have tried to get around this by rounding to 12 decimals so that small numerical differences would vanish, others just don't care (and in some methods ties really are rare). Rounding to 12 decimals may not be sufficient (issue #132) and it is never safe: two nearly identical floating point numbers can round differently. The canonical way is to add or subtract a tiny number so that values that are nearly identical are regarded as identical when using `<=`  or  `>=`. This is done with *all* permutation *P*-values in this PR. The PR also removes the now unnecessary and always unsafe `round(x, 12)`  from these functions. `CCorA`  already used this strategy and was not changed.

In all cases we use `EPS = sqrt(.Machine$double.eps)` which is about 1.5e-8 in my PC. PR also uses differences (`z >= x - EPS`) which may not be sufficient with very large real numbers. However, the new strategy changes no test result in `R CMD check`.

This PR fixes issue #132 and also gives a better fix to issue #120 than the original one.

### DIFF
--- a/R/adonis.R
+++ b/R/adonis.R
@@ -3,6 +3,7 @@
              contr.unordered="contr.sum", contr.ordered="contr.poly",
              parallel = getOption("mc.cores"), ...)
 {
+    EPS <- sqrt(.Machine$double.eps) ## use with >= in permutation P-values
     ## formula is model formula such as Y ~ A + B*C where Y is a data
     ## frame or a matrix, and A, B, and C may be factors or continuous
     ## variables.  data is the data frame from which A, B, and C would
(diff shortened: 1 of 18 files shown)


gavinsimpson:
This looks good. Merging now.

Issue #127, opened 2015-08-04
Make `scaling` more user friendly


gavinsimpson:
I am forever forgetting which way round the scaling numbers go for "site"- or "species"-focussed scalings. The numerical concept for scalings is canonical because of Canoco (& related software) but even that has long-since dropped that convention in the main user interface.

Proposal is to allow informative, non-numeric, indicators for the type of scaling within *all* functions that support a `scaling` argument and where such a change makes sense.

Thoughts on the initial idea before we speak to specifics of how to implement and what to use for the scaling *strings*?

Places we need to change this are, at a minimum:

 * [x] R/biplot.rda.R
 * [x] R/plot.cca.R
 * [x] R/plot.prc.R
 * [x] R/points.cca.R
 * [x] R/predict.cca.R
 * [x] R/predict.rda.R
 * [x] R/scores.cca.R
 * [x] R/scores.rda.R
 * [x] R/summary.cca.R
 * [x] R/summary.prc.R
 * [x] R/text.cca.R
 * [x] R/tolerance.cca.R

jarioksa:
I agree: this is a good idea. 

Strings are easy to find for non-negative values (for instance, `"none"`, `"sites"`, `"species"`, `"symmetric"`). However, the coding got overloaded with negative scores that have different meaning for `cca` and `rda` objects. How to name those?

gavinsimpson:
To do this properly, I think we need to introduce a new argument and treat linear, unimodal, others(?) separately. For unimodal methods negatives are so-called "Hill's scaling" in Canoco. For linear methods Canoco divides by the variable standard deviation, which is also what we do is it not?

If so, `scores.cca` could have argument `hill = FALSE`, and new `scores.rda` could have `standardise = FALSE` to invoke the negative scores. We'd ignore these new args if `scaling` is supplied as a length one integer.

gavinsimpson:
@jarioksa Any thoughts on suggested new arguments `hill = FALSE` and `standardise = FALSE` (or `scale = FALSE`) for `scores.cca` and `scores.rda` respectively?

jarioksa:
[...]

PR #137, opened 2015-08-30
User friendly scaling

gavinsimpson:
This PR implements changes discussed in #127 to allow for user-friendly specification of scaling types in ordinations.

### DIFF
--- a/R/biplot.rda.R
+++ b/R/biplot.rda.R
@@ -12,9 +12,10 @@
         NextMethod("biplot", x, ...)
 }
 
-`biplot.rda` <- function(x, choices = c(1, 2), scaling = 2,
-                       display = c("sites", "species"),
-                       type, xlim, ylim, col = c(1,2), const, ...) {
+`biplot.rda` <- function(x, choices = c(1, 2), scaling = "species",
+                         display = c("sites", "species"),
+                         type, xlim, ylim, col = c(1,2), const,
+                         correlation = FALSE, ...) {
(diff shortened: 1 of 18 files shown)


gavinsimpson:
Re `head(....)` this was my mistake - I intended to select a single set of scores which we could `head()` but I didn't intend to leave these `head()` bits in the code once you were happy to merge. I guess I got distracted when i made the PR and didn't include a note to that effect and that I was hoping you take a look through the code.

One thing that I haven't addressed at all here is that several methods store the `scaling` and some other even do things with this, and expect a numeric, such as one of the `print` methods. At the moment I haven't touched this aspect of the code base; once the string scalings are converted to the appropriate numeric, **vegan** works the same as before. We should revisit this but probably in another PR later so that we present user-friendly scaling names to the user when printing info.

jarioksa:
I guess you mean `print.summary.cca()`  which indeed tries to be user-friendly and give a verbal explanation how sites and species were scaled. I think we may internally keep the numeric coding which combines all things you did together -- except the value of `const` in `scores.rda()`.

I think this is good now. Passed my basic tests. There are many changes all over, and if some things were overlooked, we'll find them in wider use and fix them as they come along. 

`scores`  as such is very old (it was first in **vegan** 1.4-0 in 2002), and some of its methods have collected a huge load of arguments over the default method. This makes the general usage sometimes difficult. We have relied on `...` with these.

PR #138, opened 2015-09-03
Multiple plot parameters in ordihull-type functions

j-tweed:
Dear Jari,

I wrote to you about 18 months about this idea. Sorry it has taken me so long to implement it! From my earlier email:

> One task I often use `vegan` for is plotting ordinations showing multiple ordihulls. I often like to give these hulls different colours and/or line types. I am rather lazy and to save myself some typing I altered the original ordihull function to allow the specification of multiple values for col, lty, lwd and border – recycling as necessary over the levels of ‘groups’. (thus avoiding multiple calls to ordihull and the show.groups argument).

In response to your questions from back then, the arguments are applied in the order of `names(table(groups))`, and will also (without warning) count groups of 1 element which do not get ellipses/hulls/spiders plotted for them. Vectors are recycled if shorter than `length(table(groups))` and I've tried to ensure that parameter values are taken from `par()` when not specified by the user.

This is how it looks:

```r
data(dune)
data(dune.env)
ord <- metaMDS(dune)

old <- par(mfrow=c(1,2))
ordiplot(ord, type="n")
ordiellipse(ord, groups = dune.env$Use, draw="polygon",
 col=c("red", "blue", "green"), border=NA)

ordiplot(ord, type="n")
ordihull(ord, groups = dune.env$Use, draw="polygon",alpha=200,
 col=c("lightgrey", "lightblue", "lightgreen"), lty=c(2,1,3), lwd=c(2,1,2))

par(old)
```

image

`show.groups` still works (hopefully) as expected

```r
ordiplot(ord, type="n")
ordihull(ord, groups = dune.env$Use, draw="polygon",alpha=200, col=c("lightgrey", "lightblue"),
 lty=c(2,1), lwd=2, show.groups = c("Haypastu","Hayfield"))
```

image

I look forward to your feedback. Also since this is my first 'pull request' - please let me know if I have done anything wrong procedure-wise.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,9 @@
+# History files
+.Rhistory
+
+# Example code in package build process
+*-Ex.R
+
+## ignore backup files
+*~
+
--- a/R/ordiellipse.R
+++ b/R/ordiellipse.R
@@ -32,6 +32,7 @@
     inds <- names(table(groups))
     
     # fill in graphical vectors with default values if unspecified and recycles shorter vectors 
+    col.new <- border.new <- lty.new <- lwd.new <- NULL
     for(arg in c("col","border","lty","lwd")){
       tmp <- mget(arg,ifnotfound=list(NULL))[[1]]
       if(is.null(tmp)) tmp <- ifelse(suppressWarnings(is.null(par(arg))), par("fg"), par(arg))
--- a/R/ordiellipse.R
+++ b/R/ordiellipse.R
@@ -2,7 +2,7 @@
     function (ord, groups, display = "sites", kind = c("sd", "se"),
               conf, draw = c("lines", "polygon", "none"),
               w = weights(ord, display), col = NULL, alpha = 127,
-              show.groups, label = FALSE,  ...)
+              show.groups, label = FALSE, border=NULL,lty=NULL, lwd=NULL, ...)
 {
     weights.default <- function(object, ...) NULL
     kind <- match.arg(kind)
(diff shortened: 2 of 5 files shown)


jarioksa:
Thanks for this. @gavinsimpson or I are going to have a look at this. I'll be busy over the weekend, but Gav may be faster to comment.

j-tweed:
Sorry, I should have run the checks locally before initiating the pull request. I've learned my lesson!

jarioksa:
One tricky thing in `ordiellipse`  and `ordihull`  is that the `col`  argument defines either line colour (when `draw = "lines"`) or the fill colour (when `draw = "polygon"`). I think the default fill colour should be taken from `par("bg")` (with me `"transparent"`) whereas line colour should be `par("fg")` (with me `"black"`). This at least is the current practice. See the following with your new code:
```
ordiellipse(ord, groups = dune.env$Use, draw="polygon")
```
which draws non-transparent (`alpha` ignored) black blobs. The old function draws unfilled polygons (or filled with `"transparent"`). Actually, I think that default should always be `"transparent"` instead of `par("bg")` which can cover other symbols if `bg`  is reparametrized.

j-tweed:
I have modified `ordiellipse` and `ordihull` to supply the requested defaults for the colours of lines and fills when not specified by the user. The solution involved creating copies of `lwd`, `lty`, `col` and `border` - I hope this doesn't lead to any other unexpected behaviour.

`alpha` was also ignored when `col` was not supplied by the user, and given that the new default is to have `"transparent"` fill I didn't see a need to alter this.

Issue #129, opened 2015-08-08
Add environmental data for BCI dataset?


gavinsimpson:
In the ESA workshops we use the environmental data for the BCI plots along with the BCI data provided by **vegan**. Consider adding these data as a data object in the package.

jarioksa:
Meaningful data are available? If so, please add them.

jarioksa:
Solved with PR #141

PR #141, opened 2015-09-18
implements wish of #129 adding the BCI site information as BCI.env

gavinsimpson:
This adds the site info data from the supplementary materials. It includes some categorical variables, amount of precipitation, elevation, plus UTM coordinates.


gavinsimpson:
This is a trivial addition but would be good to get it into 2.3-1 if we can? This might cause a conflict with BiodiversityR, which also contains the same data set, but the correct solution would be for Roland to remove it from there if it was too troublesome.

jarioksa:
Roeland suggested moving several functions from **BiodiversityR** to **vegan**, but I've been too lazy to tackle the task. Having these data could also be nice. Is there really a problem of name clash if the data are identical? Never thought of this. It is clear that we cannot create conflicts with **BiodiversityR** in a minor release with this time table.  If no conflict is generated, just go on.

gavinsimpson:
The conflict is not serious; from `?data` the first matching version will be loaded, which as far as I call tell shouldn't cause any problems. The conflict is really just a name clash. I'll try to run the reverse dependency checks myself to see.

jarioksa:
Fine. I was thinking it could be something like that, and if identically named files are identical, that's OK. Just go on, but please tell Roeland so that he can adapt **BiodiversityR**.

gavinsimpson:
Good catch on Pike vs Pyke. Will fix. On the documentation, I have done that already in this PR, see: https://github.com/vegandevs/vegan/pull/141/files#diff-01d3a141e435e9bac4b8ac92280d7d88L60

jarioksa:
Merged after some small edits. I had not earlier had these, because these are pretty useless data. The only useful piece of information really are the coordinates that were calculated in 'man/BCI.Rd'. The other data are not very useful: Precipitation constant 2530mm, Elevation constant 120m, Geology constant Tb. The only variable with some variability is Age category where one plot is in `c2` and other 49 are in `c3`. I just wonder how you used these data in teaching? Having explicit coordinates is useful, though.

gavinsimpson:
Re teaching we've used the coordinates mainly so having those in an object is handier than creating them each time. Whilst rest of data is pretty useless it seemed more hassle to delete these records from the original data file than include them document the handful of variables.

jarioksa:
It seems you want this to get to 2.3-1 release. Have `tools::check_packages_in_dir()` ed this? Mainly against **BiodiversityR** which also has the same data set.

gavinsimpson:
Not successfully (reinstalled both laptop and workstation recently and not fully set-up the way I had them before) and I can't quite coax `tools::check_packages_in_dir()` to work; BiodiversityR keeps failing `check` due to missing dependencies but I have these installed so I guess I have a problem finding the library for my R when it invokes these checks.

I'll grab the sources for BiodiversityR now and build it locally against an installation of vegan 2.3-1 to check this.

gavinsimpson:
We're good to go on including this in 2.3-1. No problems checking BiodiversityR with `--as-cran`. I'll cherry pick my commits and your edits into the 2.3-1 branch later this evening or early tomorrow - depends how quickly I get the kids to bed :-)

gavinsimpson:
@jarioksa This is now merged into the cran-2.3 branch. I'll email Roland now and let him know about the new data set.

PR #150, opened 2015-12-03
Fix partial matching in load hook

richfitz:
As is, this triggers a warning on package load when using
`options(warnPartialMatchArgs=TRUE)`.


jarioksa:
Thanks. Merged.

PR #157, opened 2016-02-04
changed names in print method to clarify

EDiLD:
Changed names in print method to clarify output.
Should be now identical to documentation.

### DIFF
--- a/R/simper.R
+++ b/R/simper.R
@@ -126,8 +126,8 @@
 {
     if (ordered) {
         out <- lapply(object, function(z)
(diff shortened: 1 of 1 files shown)


jarioksa:
Looks fine: merging.

EDiLD:
Jupp, was changed because a user was confused.

Thanks.

Gesendet mit AquaMail für Android
http://www.aqua-mail.com


Am 4. Februar 2016 18:01:37 schrieb Jari Oksanen <notifications@github.com>:

Issue #161, opened 2016-04-14
Interest in adding r2-adjusted for CCA?


dmcglinn:
Peres-Neto et al (2006) developed a permutation based method for computing an adjusted r2 for cca. I have code that can accomplish this that I would be happy to contribute to the package if there was any interest in adding it in. This would add functionality to the existing function `RsquareAdj`. 

Peres-Neto, P., P. Legendre, S. Dray and D. Borcard. 2006. Variation partitioning of species data matrices: estimation and comparison of fractions. Ecology 87, 2614–2625.

gavinsimpson:
Hi Dan,

Yes, we'd like to see this in vegan, if only for consistency. Jari and I discussed this on and off in relation to some of my vegan slide decks but I never found the time to read the paper or do the implementation.

How do you want to go about this? If you don't want or have time to integrate this directly into vegan I'd be happy to take your code and get in into the package. Or you could do this yourself. I don't mind either way, whatever works best for you.

Thanks for this,

G

jarioksa:
We really want to see this in vegan. Like Gav, I never had time to check how to do this in **vegan**, but we'd really like to have it there. With the latest changes, we also have `RsquareAdj` for db-RDA, and this would remove the last exception from constrained methods.

gavinsimpson:
Closed via bbaf30d

PR #162, opened 2016-04-14
This PR implements an adjusted R2 calculation for cca

dmcglinn:
Based on conversations in #161 I decided to provide code for implementing the permutation based approach of adjusting the R2 calculation for cca that was developed by Peres-Neto et al 2006. There are currently some limitations of my implementation (see commit log). I'm happy to work to overcome these limitations ironed out so that this method can be more fully realized, but I'll likely need some suggestions for how to do this.

### DIFF
--- a/R/RsquareAdj.R
+++ b/R/RsquareAdj.R
@@ -35,10 +35,10 @@
 
 ## cca result: no RsquareAdj
 RsquareAdj.cca <-
-    function (x, nperm = 1000) 
+    function (x, permutations = 1000, ...) 
 {
     r2 <- x$CCA$tot.chi / x$tot.chi
-    p <- permutest(x, nperm)
+    p <- permutest(x, permutations, ...)
     radj <- 1 - ((1 - r2) / (1 - mean(p$num / x$tot.chi)))
     list(r.squared = r2, adj.r.squared = radj)
 }
--- a/R/RsquareAdj.R
+++ b/R/RsquareAdj.R
@@ -35,11 +35,30 @@
 
 ## cca result: no RsquareAdj
 RsquareAdj.cca <-
-    function(x, ...)
+    function(x, nperm, print_progress=TRUE, ...)
 {
-    R2 <- x$CCA$tot.chi/x$tot.chi
-    radj <- NA
-    list(r.squared = R2, adj.r.squared = radj)
+    r2 <- x$CCA$tot.chi/x$tot.chi
+    n <- nrow(x$CCA$u)
+    rand_r2 <- rep(NA, nperm)
+    Y_string <- as.character(x$terms[[2]])
+    Y <- eval(parse(text=Y_string))
+    Y_name <- strsplit(as.character(x$call[2]), ' ~ ')[[1]][1]
+    x$call[2] <- sub(Y_name, 'Y_rand', x$call[2])
+    for (i in 1:nperm) {
+        Y_rand <- Y[sample(n), ]
+        cca_rand <- eval(parse(text=paste(x$call[1], '(',x$call[2], 
+                                         ', data=', x$call[3], ')', 
+                                         sep='')))
+        rand_r2[i] <- cca_rand$CCA$tot.chi / x$tot.chi
+        if (print_progress) {
+            if (i %% 100 == 0)  
+                print(paste('perm:', i, 'adj.r.squared:', 
+                            round(1 - ((1 - r2) / (1 - mean(rand_r2, na.rm=T))),3)))
+        }
+    }
+    # Eq 5 Peres-Neto et al. 2006
+    radj <- 1 - ((1 - r2) / (1 - mean(rand_r2)))
+    list(r.squared = r2, adj.r.squared = radj)
 }
 
 ## Linear model: take the result from the summary
--- a/R/RsquareAdj.R
+++ b/R/RsquareAdj.R
@@ -42,8 +42,7 @@ RsquareAdj.cca <-
     rand_r2 <- rep(NA, nperm)
     Y_string <- as.character(x$terms[[2]])
     Y <- eval(parse(text=Y_string))
-    Y_name <- strsplit(as.character(x$call[2]), ' ~ ')[[1]][1]
-    x$call[2] <- sub(Y_name, 'Y_rand', x$call[2])
+    x$call[2] <- sub(Y_string, 'Y_rand', x$call[2])
     for (i in 1:nperm) {
         Y_rand <- Y[sample(n), ]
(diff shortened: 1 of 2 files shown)


jarioksa:
It seems that I need to do some reading. My impression was that Peres-Neto's method was more complicated, but I never tried to dive through the details of calculation. Your implementation is very direct and simple. Actually, it could be written more briefly using `vegan:::permutest.cca`:
```R
RsquareAdj.cca <-
    function (x, nperm = 1000) 
{
    r2 <- x$CCA$tot.chi/x$tot.chi
    p <- permutest(x, nperm)
    radj <- 1 - ((1-r2) / (1 - mean(p$num/x$tot.chi)))
    list(r.squared = r2, adj.r.squared = radj)
}
```
This is also faster and easier to maintain. However, these approaches should be verified against Peres-Neto's paper. Do you have any test case?

BTW, the main usage of `RsquareAdj` is in embedded usage in functions like `ordiR2step`  and `varpart`. When I enabled your code in `ordiR2step` I got an error:
```R
Error in svd(Xbar, nu = 0, nv = 0) : a dimension is zero
> traceback()
10: stop("a dimension is zero")
9: svd(Xbar, nu = 0, nv = 0)
8: cca.default(d$X, d$Y, d$Z)
7: cca.formula(Y_rand ~ 1, data = dune.env)
6: cca(Y_rand ~ 1, data = dune.env) at <text>#1
5: eval(expr, envir, enclos)
4: eval(parse(text = paste(x$call[1], "(", x$call[2], ", data=", 
       x$call[3], ")", sep = ""))) at RsquareAdj.R#48
3: RsquareAdj.cca(update(object, fla))
2: RsquareAdj(update(object, fla)) at ordiR2step.R#67
1: ordiR2step(m0, scope = formula(m))
```
I think this happens because the function tries to evaluate an empty model with no constrained component  `cca(Y ~ 1)`.

dmcglinn:
Hey @jarioksa thanks for the feedback. The `permutest` implementation is a great suggestion. I'll integrate that. Yea the approach is pretty simple. See equ 5 in the Peres-Neto et al 2006 paper. Do you basically envision several `if` statements within this function to catch all the special cases of different ways a cca model could be specified (e.g., the empty model example) or is their a cleaner and more direct way to go?

dmcglinn:
Also I don't have a clean test case to compare with the Peres-Neto paper. They did provide some matlab code and an executable. I'll do some manual comparison their to see if the values match.

jarioksa:
When I checked the `permutest` implementation, it  passed the `ordiR2step` case: it seems that `permutest.cca` handled smoothly missing `CCA` component. 

We use `RsquareAdj` in `ordiR2step` and in `varpart`, and both currently check the input and refuse to work with `cca` results. Handling of  `cca` results should be accepted, and probably also an example added, and possibly also a test case in `tests/vegan-tests.R` to catch regressions.

dmcglinn:
Nice ok well i'll report back on my manuel testing foray

jarioksa:
@gavinsimpson : in permutation tests we add observed statistic with permutations, and therefore we have, say, 999 permutations + one observations. What about permutation-based R²? Should we have the observed R² there, too?

dmcglinn:
Peres-Neto did not specify one way or another on if the observed R2 should be include. I would lean towards not including because if the number of permutations is small this will bias the result and the permuted R2 represent the explained variance of random variables where as in a permutation test it represents potentially observable shufflings of the data which have slightly different interpretations I think. 

As a test, I used the executable program archived here: http://www.esapubs.org/archive/ecol/E087/158/suppl-3.htm  to compute the cca r2 and adjusted r2 for the `varespec` and `varechem` datasdets. Here is the output from that analysis: 
[cca_r2.txt](https://github.com/vegandevs/vegan/files/221171/cca_r2.txt)

This matches r2 and adjusted r2 computed for my suggested code to the second decimal point with 1000 permutations.

gavinsimpson:
I agree with Dan's assessment re including observed R2 in the permutation distribution. Pedro and co are not using the permutations in a hypothesis testing mode (looking for evidence to reject a null) but are simply generating a null distribution from which to compute an estimator for p/(n - 1). Though the distinction is small.

dmcglinn:
Hey @jarioksa and @gavinsimpson I think the function `RsquareAdj.cca` is ready to go now. I implemented the version that @jarioksa suggested and I updated the man file accordingly.

R/RsquareAdj.R
```diff
@@ -35,11 +35,12 @@
 
 ## cca result: no RsquareAdj
 RsquareAdj.cca <-
-    function(x, ...)
+    function (x, nperm = 1000)
```
gavinsimpson: Function definition needs to have `...` as an argument because the generic has this argument.

For consistency, `nperm` should be renamed to `permutations` as per `anova.cca` etc.

R/RsquareAdj.R
```diff
@@ -35,11 +35,12 @@
...
 ## cca result: no RsquareAdj
 RsquareAdj.cca <-
-    function(x, ...)
+    function (x, nperm = 1000) 
 {
-    R2 <- x$CCA$tot.chi/x$tot.chi
-    radj <- NA
-    list(r.squared = R2, adj.r.squared = radj)
+    r2 <- x$CCA$tot.chi / x$tot.chi
+    p <- permutest(x, nperm)
```
gavinsimpson: How about passing `...` to `permutest()` here? That way people can call this with `parallel` or other functions should they need to speed this up.

jarioksa:
Yes, we should be able to pass `parallel` to `permutest.cca`: we need it in particular when `RsquareAdj.cca` is embedded in other functions and called repeatedly.

dmcglinn:
done

gavinsimpson:
Thanks for handling this @dmcglinn. Merged.

dmcglinn:
It was great working with both of you thanks for your encouragement and guidance!

gavinsimpson:
Cheers @dmcglinn and thanks again for this - @jarioksa and I have spoken about this on and off for a while but we never got round to to doing anything about it. Now I can update my slides ahead of the Vegan workshop at ESA to say this is done!

jarioksa:
Thanks @dmcglinn . I have now allowed `cca` in `ordiR2step` in a separate branch (which may need some scrutiny before the release). However, it seems that we still can't implement variation partition (`varpart`) with CCA: we use there a minimal `simpleRDA2` function followed with `RsquareAdj.default`, and the current `RsquareAdj.cca` cannot be pluggled in. (It is surprisingly trivial to estimate unadjusted R squared for CCA using `simpleRDA2, but adjusted R-squared needs more coding.)

Issue #128, opened 2015-08-08
Allow more control over colours / groups for plot.betadisper


gavinsimpson:
Several of the drawn components of the `plot.betadisper()` output are hardcode. Allow users to specify vectors of colours for these elements.

jarioksa:
Fine! Please do this.

CarlyRae:
Hello,

I requested this during the Vegan workshop at ESA. I am trying to figure out which pch characters relate to which factor level in the plot.betadisper. 

It would be great to have some additional control over the plot that is created. 

For the time being could you tell me how to find out which pch is which group?

For instance, I did this 

mod2_ele <- betadisper(uni_ele, ele_group)
anova(mod2_ele)
plot(mod2_ele)

str(mod2_ele)

#### From str I can see the group order, but where can I find how the pch are specified? 

 $ group    : Factor w/ 4 levels "1000","700","800",..: 4 1 4 4 1 1 1 4 1 3 ...

Thanks and would love to see more features with this plotting function. The figures are informative.

gavinsimpson:
@CarlyRae OK - I'll tackle this ahead of the release vegan version 2.4-0. We've only just started talking about this and there are a lot of candidates for inclusion in 2.4-0 so it might not be on CRAN for a while, but try to get some time to implement something for `plot.betadisper()` in the next week or two.

CarlyRae:
Great, I look forward to seeing what you develop.

jarioksa:
I think we are done with this after @gavinsimpson committed baea831

PR #165, opened 2016-04-16
pch= and col= args added to plot.betadisper()

friendly:
In response to issue #128 and for my needs, I added basic capability to set pch= and col=
for the points plotted for groups.  There are still quite a few other graphical parameters hardcoded,
but I resisted the temptation to go further in this round.


gavinsimpson:
Thanks @friendly. I'm just running checks locally before merging this.

Issue #166, opened 2016-04-16
add veganCovEllipse() as an option to plot.betadisper()


friendly:
Hulls are useful for one type of display, but covariance (data) ellipses provided by
veganCovEllipse are also useful.  It would not be hard to add this via an option
ellipse=T/F, as an alternative to hull=T/F

gavinsimpson:
Sure, should be relatively easy to implement.

gavinsimpson:
I've added an option to do this, though I'm not 100% convinced I have this right yet. If you have any comment or input, see PR #167

gavinsimpson:
Closed via #167

friendly:
The most recent version is now extremely nice.  thanks for the work on this./

jarioksa:
@gavinsimpson : When I look at the graph in [the bottom of the heap](http://www.fromthebottomoftheheap.net/2016/04/17/new-plot-default-for-betadisper/), it seems that `plot` with ellipses uses centroids whereas the hulls (and the analysis) used spatial medians. It may be difficult to reconcile covariance ellipses (L2 norm) and median statistics (L1 norm).

friendly:
When I look at the plots on that page, it seems to me that the centers (group labels) are in the same positions in both the plot with hulls and that with ellipsoids.  If you're referring to the fact that
the centers from which the segments are drawn are not at the ellipse centers (L2 norm), I don't find this a problem, as long as it is understood that the covariance ellipse is just a visual summary of the bivariate scatter of the points shown under normal theory in the space of the principal coordinates. 

Ellipses may not be that useful for some distance metrics, but they certainly are for Euclidean distances.

jarioksa:
It's not about the metric (Euclidean or non-Euclidean), but it is about the way `betadisper` makes its calculations: we have argument `type = c("median", "centroid")`, where the default is `"median"` which also is displayed as the centroid of hulls. However, ellipses will always show `"centroid"` even when the analysis is not based on them. It seems we *could* use the calculated centroids, as `cov.wt` takes a `center` argument which can be a user-supplied vector of centres used in calculating covariances.

gavinsimpson:
@friendly the problem is that I didn't plot the ellipsoid centre at all; the code just plots the `centroid` component of the analysis, which by default is the cluster *median* (I'd forgotten we'd changed the default back when we were discussing bias adjustments in a previous update).

So two options:

1. draw the ellipse centre if user asks for ellipses as the method used `type = "median"`,
[...]

PR #167, opened 2016-04-16
add option to draw confidence ellipses to group centroids

gavinsimpson:
As per feature request of #166 I implemented a first pass at this based on code from `ordiellipse()`. This is much simpler though as I only needed the code to draw ellipses as lines, not polygons and `plot.betadisper` draws these things in a loop, group by group, which further simplifies the code.

I'm not quite convinced I have this correct; needs to use `cov.wt()` as per `ordiellipse()` but I don't have `weights` for the PCoA used in `betadisper()` so I just set this to a vector of `1`s, again following `ordiellipse()`.

@jarioksa would you mind taking a look at this?

### DIFF
--- a/R/plot.betadisper.R
+++ b/R/plot.betadisper.R
@@ -1,10 +1,26 @@
 `plot.betadisper` <- function(x, axes = c(1,2), cex = 0.7, pch = seq_len(ng),
                               col = NULL, hull = TRUE,
+                              ellipse = FALSE, ellipse.type = c("sd","se"),
+                              ellipse.conf = NULL,
                               ylab, xlab, main, sub, ...)
 {
     localAxis <- function(..., col, bg, pch, cex, lty, lwd) axis(...)
     localBox <- function(..., col, bg, pch, cex, lty, lwd) box(...)
     localTitle <- function(..., col, bg, pch, cex, lty, lwd) title(...)
+    Ellipse <- function(scrs, type, conf, col, lty, lwd) {
+        mat <- cov.wt(scrs, wt = rep_len(1, nrow(scrs)))
+        if (mat$n.obs == 1)
+            mat$cov[] <- 0
+        if (type == "se") {
+            mat$cov <- mat$cov * sum(mat$wt^2)
+        }
+        xy <- if (mat$n.obs > 1) {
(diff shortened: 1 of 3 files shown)


R/plot.betadisper.R
```diff
@@ -1,10 +1,26 @@
...
                               col = NULL, hull = TRUE,
+                              ellipse = FALSE, ellipse.type = c("sd","se"),
+                              ellipse.conf = NULL,
                               ylab, xlab, main, sub, ...)
 {
     localAxis <- function(..., col, bg, pch, cex, lty, lwd) axis(...)
     localBox <- function(..., col, bg, pch, cex, lty, lwd) box(...)
     localTitle <- function(..., col, bg, pch, cex, lty, lwd) title(...)
+    Ellipse <- function(scrs, type, conf, col, lty, lwd) {
+        mat <- cov.wt(scrs, wt = rep_len(1, nrow(scrs)))
```
jarioksa: This can be simplified slightly: weights are not needed if you have unweighted covariance, but you can write
```R
mat <- cov.wt(scrs)
```
and then change line 15 below into
```R
if (type == "se") {
    mat$cov <- mat$cov / mat$n.obs
}
```
Using `cov.wt` is practical, though, because it gives you both the covariances and the centroids.

friendly:
This seems good, just looking at the code, though it isn't clear why cov.wt is used; does no harm.
I'll give it a try with my running example.

friendly:
Fixed a bug with conf, and I added ellipse.lwd control.  I'll issue a pull request from my fork.

Here is the example I'm using:

    dst <- dist(iris[,1:4])
    iris.bd <- betadisper(dst, iris$Species)
    labs <- paste("Dimension", 1:4, "(", 
                  round(100*iris.bd$eig / sum(iris.bd$eig), 2), "%)")

    plot(iris.bd, cex=1, pch=15:17,
         main="Iris data: MDS coordinates", cex.lab=1.25,
         xlab=labs[1], ylab=labs[2],
         hull=FALSE, ellipse=TRUE, ellipse.conf=0.68, ellipse.lwd=2)

In my context, I prefer using data ellipses to convex hulls, so this is a welcome addition.  Thanks
for acting on my request.

PR #168, opened 2016-04-17
re: # 167--fixed conf bug, added ellipse.lwd in plot.betadisper.R

friendly:
As noted in the title.


gavinsimpson:
Thanks for spotting the conf bug. I've gone a different route for parameters for the hulls/ellipses; I am using the standard pars for this, having added `lty` and `lwd` that affect the hulls and ellipses only.

So I'm going to close this and apply the fix to the confidence interval separately.

friendly:
That looks even better.  For such graphs, sometimes one want to emphasize the dispersion of
the points, but sometimes the bounding ellipses/hulls.  

FYI: in the `heplots` package on R-Forge, https://r-forge.r-project.org/projects/heplots/ there is now a fairly general `covEllipses()` that provides a few more features.

PR #169, opened 2016-04-17
Improve points & text orditkplot methods

gavinsimpson:
This PR makes the `points()` and `text()` methods for class `"orditkplot"` more broadly useful by using the stored plot parameters for things like `pch`, `cex`, `bg` etc. Rationale for this is that plots are often built up layer by layer and and it seems odd and inconsistent to not use the information stored in the `orditkplot` object if the user doesn't supply any of those parameters.

@jarioksa any objections to this? The existing code/documentation was pretty explicit about ignoring those parameters.

### DIFF
--- a/R/points.orditkplot.R
+++ b/R/points.orditkplot.R
@@ -1,6 +1,5 @@
-`points.orditkplot` <-
-    function(x, ...)
-{
-    points(x$points, ...)
+`points.orditkplot` <- function(x, pch = x$args$pch, cex = x$args$pcex,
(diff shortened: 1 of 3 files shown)


jarioksa:
I don't remember what I was thinking, but probably I was thinking something when I drafted these functions in 2008. Possibly my idea was to give the user an option to over-rule my design decisions. The suggested change retains that right but is an improvement. NB., `orditkplot` method functions are also used for `ordipointlabel` objects. The suggested functions seem to work there, too. I'll merge this.

PR #173, opened 2016-05-19
Exposing method argument in adipart/multipart/hiersimu

psolymos:
@jarioksa suggested to expose `method` argument passed to `oecosimu` in the functions `adipart`, `multipart`, and `hiersimu`

> I think it could be useful to expose the null model argument in adipart and hiersimu. 
> You can pass it as method, but it seems that the user knows the vegan design very well to realize 
> that this is the way to pass null model. In raupcrick() I do have explicit passing just to make 
> users aware of this possibility.


jarioksa:
This looks fine. Is this ready for merging, or do you still want to check something?

Issue #164, opened 2016-04-16
Binding multiple `simmat` objects together


psolymos:
@jarioksa and I have discussed the the possibility of adding a function for binding multiple `simmat` objects. This is useful because:

* **vegan** has no option for stratification in null models,
* and parallel computing often returns a list of `simmat` objects that need to be coerced/bound.

For this end I created the `bind-simmat` branch and added the `smbind` function and its documentation. The naming refers to **s**im**m**at**bind**. I did not create a generic because the function can take multiple object classes as input, e.g. `smbind(x, y)` or `smbind(list(x, y))` which are the 2 common use cases I think (similarly to `AIC()`).

psolymos:
Start, end, thinning (~ts attributes), and number of simulated matrices need to be consistent in for sequential algorithms. Here is the decision tree along which the current `smbind` implementation works:

1. `MARGIN != 3`: stratification intended, all start, end, thin must equal (dimensions are checked too).
2. `MARGIN == 3`:
  1. all start, end, thin equal: adequate for null model analyses but ts attributes are set to `NA`;
  2. thin equal, start and end is consistent with subsequent samples, ts attributes are updated accordingly.

In all other cases an error is produces, unless `strict = FALSE`.

jarioksa:
For most use cases we ignore the chains in sequential methods and treat the combined result only as a 3D array. However, for some diagnostic tools we would like to have the information on chain. That is, `oecosimu` can be ignorant of chains, but `as.mcmc` and the **coda** tools should have the information. It seems that **coda** is rather picky for multiple chains: `?mcmc.list` says:

> The list must be balanced: each chain in the list must have the same iterations and the same variables.

That means that `start`, `end` and `thin` (and hence length) must be equal, but are each repeated for each chain. Such a data will be generated if a sequential model is called `chains` times in parallel. Shouldn't such cases be handled? The only function that really must be aware of this is `as.mcmc` (and perhaps `as.ts`). This could happen, e.g., so that `smbind` adds a `chains` argument if it combines sequential models with equal parallel `start`, `end` and `thin`, and then `as.mcmc` breaks results in to a `mcmc.list` when needed. I don't think we need this even for `permustats.oecosimu` or at least we have no tools of handling `permustats.oecosimu` results as sequential results from different chains. Here an example of hand-crafted `mcmc.list` object that was a result of parallel processing of `sipoo` data using `"swap"` without burnin or thinning. 
[...]

PR #175, opened 2016-05-20
Merge bind-simmat branch to master: adding new function `smbind`

psolymos:
The new function `smbind` is related to issue #164 works as expected. Examples cover the use cases and results are as expected. Also fixes some issues with start/end/thin values in `as.ts.oecosimu`, `simulate.nullmodel`, and `update.nullmodel`.


jarioksa:
I have run some tests stratifying species by taxonomy, and everything seems to work fine. The usage requires some familiarity with **R** intricacies, though. Perhaps `permatswap` or `permatfull` interface could be used to make stratifying easier. 

So far my only comment is that you should not reset RNG in the user session. This would happen with `example(nullmodel)`. I have suggested a change to this in daad496c90db4397a56efd0b85f11cbaf23473d1. This is similar as in `man/orditorp.Rd`, and it would be still better to save the state of RNG and reset to it at exit (in both cases).

It is a question of taste, but I do prefer to have real data sets in examples instead of simulated ones.

psolymos:
I have no problem with changing the matrix to `data(dune); x <- as.matrix(dune)` in `example(nullmodel)`. That would address both the seed reset issue and preference towards real data sets.

I merge this branch and follow up with the example and null model error messages as references in #177

Issue #197, opened 2016-09-15
.Call interface for nullmodel simulation


jarioksa:
I have started implementing `.Call()` interface to `nullmodel` simulation. Currently we have several compiled **C** functions and we access these with `.C()` interface as set up in `make.commsim()` and used then by `simulate.nullmodel()`. I wanted to see if these functions can be made more efficient by switching from `.C` to `.Call`. 

In `.C` interface we pass basic **C** structures, typically **R** data as (pointers to) **C** vectors. The passing of arguments and return of results happens via arguments. In `.Call` interface we pass **R** structures and can return **R** structures. I have now implemented two test cases which together cover all variants in `nestedness.c` (i.e., sequential and non-sequential models). For sequential methods I have implemented `.Call` to trial swap, and for non-sequential the `.Call` to quasiswap. The traditional interface to trial swap is:
```R
    out <- array(0L, c(nr, nc, n))
    out[, , 1] <- .C("trialswap", m = x, nr, nc, thin, PACKAGE = "vegan")$m
    for (k in seq_len(n - 1)) out[, , k + 1] <- .C("trialswap", 
        m = out[, , k], nr, nc, thin, PACKAGE = "vegan")$m
    out
```
We first create an empty 3D array `out`, then fill the first facet (third dimension, `out[,,1]`) with swapped input matrix (`m`), and then we loop over all `n` simulations and fill the next matrix with the swapped previous one with repeated `.C("trialswap", ...)`. The new interface is:
```R
 .Call("do_tswap", as.matrix(x), n, thin, PACKAGE = "vegan")
```
That's all. We pass an **R** matrix, create the 3D array within the compiled code, loop over its third facet in compiled code, and finally return the 3D array. The `do_tswap` function will call the very same `trialswap` **C** code as the old interface, but this call happens within compiled code instead of crossing **R**/**C** border.  Using `.Call` interface normally means moving a part of **R** into **C** while keeping the old **C**.

I would appreciate tests and comments on the new interface. The new interface is now available in branch `nestedness-Calls` of the devel branch of **vegan**.  The new interfaces can be defined in the `nullmodel` command or in `oecosimu`. For instance, the version above can be invoked as:
```R
nm <- nullmodel(BCI, "dotswap")
oecosimu(BCI, nestedchecker, "dotswap", thin=1000)
```
[...]

PR #201, opened 2016-09-27
Nestedness Calls

jarioksa:
This PR changes sequential `nullmodel` methods and `"quasiwap_count"` to use a single `.Call()` instead of an **R** loop with `.C()` calls as discussed in issue #197. According to fairly extensive testing this is always faster and sometimes very much faster than the old interface. With the same random number seed the results are identical to the previous interface. The actual user-interface is unchanged, and users should see no change, except in speed.

The `"quasiswap"` still uses the old **R** loop to `.C()`. There was no detectable difference in speed between interfaces. However,  `.Call("do_qswap", ...)` is ready to  accept `"quasiswap"` if so defined in the **R** code.

There are also some changes in **C** code. The RNG is clearly faster: it no longer checks the input dimensions at every call, but checking is done in **R** before calling **C** code, and a macro is used in RNG instead of a function. The `trialswap` function uses a single `switch` by matrix signature instead of tow `if`s. The functions which are not called from  **R** functions are marked as `static` and are local to the `nestedness.c` routines.    User interrupts are checked at longer intervals.


jarioksa:
This PR makes most **C** function in `nestedness.c` invisible to **R** by defining them `static void`. External package will fail if it calls these functions with `C()` interface Therefore I ran `tools::check_packages_in_dir()` for reverse dependencies with following `summary`:
```
Check status summary:
                  ERROR WARN NOTE OK
  Source packages     0    0    0  1
  Reverse depends     3    1   20 40
```
All errors were caused by dependence on missing packages (not in CRAN), and most NOTEs were caused by problems by the package itself. There was no problem that may have caused by making some **C** functions `static`.

Issue #200, opened 2016-09-26
bug in rda formula vegan 2.4-1


GZikalala:
Hi,

I have a problems with running the rda function (vegan version 2.4-1),

    vegan::rda(Y ~ X1 + X2 + X3+ X4, data= Env_1dmod)

    "Error in formula.default(object, env = baseenv()) : invalid formula"

Help will be much appreciated.

Thanks.

Gugu

gavinsimpson:
I can't replicate this locally:

```
> library("vegan")
Loading required package: permute
Loading required package: lattice
This is vegan 2.4-1
> data(dune.env, dune)
> vegan::rda(dune ~ Use + Management + Moisture + Manure, data = dune.env)
Call: rda(formula = dune ~ Use + Management + Moisture + Manure, data =
dune.env)

              Inertia Proportion Rank
Total         84.1237     1.0000     
Constrained   60.8358     0.7232   11
Unconstrained 23.2879     0.2768    8
Inertia is variance 
Some constraints were aliased because they were collinear (redundant)

Eigenvalues for constrained axes:
  RDA1   RDA2   RDA3   RDA4   RDA5   RDA6   RDA7   RDA8   RDA9  RDA10  RDA11 
22.371 15.794  7.038  4.029  3.503  2.527  2.102  1.657  0.943  0.582  0.289 

Eigenvalues for unconstrained axes:
  PC1   PC2   PC3   PC4   PC5   PC6   PC7   PC8 
6.857 5.284 3.853 2.583 2.398 1.120 0.674 0.518
```

Suggestions are to retry in a clean session, and I assume you have already loaded the package via `library("vegan")`, if not try that and see if it works. It looks like the correct formula processing code is not being dispatched and your use of `vegan::rda` might suggest you haven't loaded the package yet.

jarioksa:
I have tried several ways of abusing `rda` command, but I haven't been able to reproduce your error. Please type `traceback()` after getting the error and show us the returned path to the error.

jarioksa:
OK. Now I can reproduce this problem:
```R
library(vegan)
data(dune, dune.env)
notadataframe <- data.matrix(dune)
rda(dune ~ A1 + Moisture, notadataframe)
# Error in formula.default(object, env = baseenv()) : invalid formula
```
Is your `Env_1dmod` a `matrix` instead of a `data.frame`? What does `class(Env_1dmod)` say?

[...]

PR #202, opened 2016-10-02
data as matrix errors

jarioksa:
This PR simplifies formula manipulations in constrained ordination. It started as a response to issue #200, and also fixes those problems: now the error message is clear and tells that `data` cannot be a matrix.

The PR deletes function `ordiGetData` which was added in **vegan_1.8-4** to remove some scoping issues. The same functionality should be achieved by the current code with standard `eval()` function. 

I also noticed that the left-hand-side (community or community dissimilarities) were evaluated twice in distance-based `dbrda` and `capscale` and this could be costly. For instance, the lhs could be function `stepacross(vegdist(x))` which can be really expensive for large data sets. Now the lhs is no longer re-evaluated when it was already evaluated in distance-based methos.

### DIFF
--- a/R/adonis2.R
+++ b/R/adonis2.R
@@ -55,7 +55,8 @@
     if (missing(data))
         data <- .GlobalEnv
     else
-        data <- ordiGetData(match.call(), environment(formula))
+        data <- eval(match.call()$data, environment(formula),
+                     enclos = .GlobalEnv)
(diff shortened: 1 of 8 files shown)


jarioksa:
Passes tests, even extended tests with old scripts. The only way to see if this is a safe fix is to merge it and release so that we start getting error messages...

PR #206, opened 2016-10-11
Allow for ordicluster and related functions to return details of what was drawn and allow suppression of plotting

gavinsimpson:
It would be useful to call `ordicluster` and possibly other functions documented in `ordihull.Rd` without the side effect of drawing on the current device. In the case of `ordicluster` it would be useful to have the segment `x` and `y` coordinates (and possibly colours) that would be drawn returned.

These could then be used for custom plotting by experts or by other packages, like **ggvegan**, to allow reproducing of **vegan**'s base graphics plots in other plotting engines.

This PR is a work in progress and the first implementation of an idea along the lines of the above. This affects only `ordicluster` and form and documentation of the returned object is not fixed.

The PR currently:
- adds arg `plot` to suppress plotting
- returns the segments and colours as drawn
- refactors to do a single call to `segments`

The latter is a logical conclusion of storing all the information for plotting and should probably be adopted even if we don't end up implementing the changes to `ordicluster` & co.

### DIFF
--- a/R/ordicluster.R
+++ b/R/ordicluster.R
@@ -1,6 +1,6 @@
 `ordicluster` <-
-    function (ord, cluster, prune=0, display="sites",
-              w = weights(ord, display), col = 1, ...)
+    function (ord, cluster, prune = 0, display = "sites",
+              w = weights(ord, display), col = 1, plot = TRUE, ...)
 {
     weights.default <- function(object, ...) NULL
     w <- eval(w)
@@ -16,7 +16,8 @@
     col <- rep(col, length = nrow(ord))
     col <- col2rgb(col)/255
     nodecol <- matrix(NA, nrow(mrg) - prune, 3)
-    for (i in 1: (nrow(mrg) - prune)) {
+    seg.coords <- matrix(NA, nrow = nrow(mrg) - prune, ncol = 4)
+    for (i in seq_len(nrow(mrg) - prune)) {
         a <- mrg[i,1]
         b <- mrg[i,2]
         one <- if (a < 0) ord[-a,] else go[a,]
@@ -30,9 +31,16 @@
         colone <- if (a < 0) col[,-a] else nodecol[a,]
         coltwo <- if (b < 0) col[,-b] else nodecol[b,]
         nodecol[i,] <- (n1 * colone + n2 * coltwo)/noden[i]
-        ordiArgAbsorber(one[1], one[2], two[1], two[2],
-                        col = rgb(t(nodecol[i,])),
+        seg.coords[i, ] <- c(one[1], one[2], two[1], two[2])
+    }
+    colnames(seg.coords) <- c("x1","y1","x2","y2")
+    seg.coords <- cbind(seg.coords, col = nodecol)
+    if (plot) {
+        ordiArgAbsorber(seg.coords[,1L], seg.coords[,2L], seg.coords[,3L], seg.coords[,4L],
+                        col = rgb(nodecol),
                         FUN = segments, ...)
     }
-    invisible(cbind(go, "w"=noden))
+    out <- structure(list(scores = cbind(go, "w" = noden),
+                          segments = seg.coords), class = "ordicluster")
+    invisible(out)
 }
--- a/R/ordicluster.R
+++ b/R/ordicluster.R
@@ -1,6 +1,7 @@
 `ordicluster` <-
     function (ord, cluster, prune = 0, display = "sites",
-              w = weights(ord, display), col = 1, plot = TRUE, ...)
+              w = weights(ord, display), col = 1,
+              draw = c("segments", "none"), ...)
 {
     weights.default <- function(object, ...) NULL
     w <- eval(w)
@@ -35,8 +36,11 @@
     }
     colnames(seg.coords) <- c("x1","y1","x2","y2")
     seg.coords <- cbind(seg.coords, col = nodecol)
-    if (plot) {
-        ordiArgAbsorber(seg.coords[,1L], seg.coords[,2L], seg.coords[,3L], seg.coords[,4L],
+    ## are we plotting?
+    draw <- match.arg(draw)
+    if (isTRUE(all.equal(draw, "segments"))) {
+        ordiArgAbsorber(seg.coords[,1L], seg.coords[,2L],
+                        seg.coords[,3L], seg.coords[,4L],
                         col = rgb(nodecol),
                         FUN = segments, ...)
     }
--- a/man/ordihull.Rd
+++ b/man/ordihull.Rd
@@ -34,7 +34,8 @@ ordispider(ord, groups, display="sites", w = weights(ord, display),
 	 spiders = c("centroid", "median"),  show.groups, 
          label = FALSE, col = NULL, lty = NULL, lwd = NULL, ...)
 ordicluster(ord, cluster, prune = 0, display = "sites",
-            w = weights(ord, display), col = 1, plot = TRUE, ...)
+            w = weights(ord, display), col = 1, draw = c("segments", "none"),
+            ...)
 \method{summary}{ordihull}(object, ...)
 \method{summary}{ordiellipse}(object, ...)
 ordiareatest(ord, groups, area = c("hull", "ellipse"), kind = "sd",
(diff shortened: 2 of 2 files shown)


jarioksa:
`ordihull` and `ordiellipse`  use argument `draw = "none"` for the same effect as you suggest `plot = FALSE`. Would it make sense to be consistent? (The `draw = "none"` argument was provided to allow significance tests such as `ordiareatest` without repeated & slow drawing.)

jarioksa:
I found a buglet in the `ordicluster` code: it was in the original code, and it persisted through your edits. The invisible output object (internal name `go`) should contain the coordinates of internal points were cluster branches are merged, but last `prune + 1` rows had the coordinates of observed ordination scores (external points). The effects are visible if you want to add the internal points to the graph `plot` (old code) or `plot(scores())` (your new code), and some of the ordination points will be overwritten. I have fixed this in commit 00bf351997f3c722c58cf3c948d534a7314cf97a in my tree (you can merge this to your tree).

gavinsimpson:
Thanks for spotting and fixing the bug in the original code.

As for `draw`, I'm conflicted. In base R `plot = FALSE` is normally used for suppressing plotting (see eg `boxplot` or `hist` as examples I was thinking about when adding the argument), but I also see that `draw` as used in `ordihull` and `ordiellipse` needs to handle more options than a simple yes/no.

This might be an opportunity to rethink `draw` slightly; how would you feel about `draw` being used to select *what* is drawn but `plot` being used to suppress plotting? For backwards compatibility we would probably want to initially retain the `draw = "none"` option and allow it to override a `plot = TRUE` default, but issue a message/warning that use of `draw = "none"` is deprecated, with a view to removing the `"none"` option at some future date.

Thoughts?

jarioksa:
Yes, I knew about the `plot = FALSE` conventions in some functions, but it was not the only convention in **R**: think about `plot(..., type = "n")` or `plot(..., type = "none")` (both are used). We had `draw` argument for the kind of plot (`lines` or `polygon`, *e.g.*), and when I needed a non-drawing alternative for `ordiareatest`, using the existing `draw` was a natural choice. Using `type = c("lines", "polygon", "none")`  would have been more consistent with basic **R** from the beginning, though. My main point was that `draw` was an existing argument in `ordi*` functions and could be reused.

gavinsimpson:
I see `type = "n"` as different to what `plot = FALSE` does; for a start `type = "n"` certainly draws something. I'm not familiar with a `type = "none"` example (unless that evaluates to `type = "n"`?)

So, yes, I get your main point. And yes I had missed that `draw` was already there. I guess what I'm asking about now is whether we stick with `draw` or rationalise the current behaviour to have both `draw` and `plot`, which do fundamentally different things.

Which would you prefer? change meaning of `draw` slightly and add `plot = FALSE` or just harmonise on `draw` even if this is a little different to other base R functions?

jarioksa:
`plot = FALSE` suits me, but that means a bigger change: all other functions having now `draw = "none"` should be changed, and so should be changed functions that call these functions with `draw = "none"` (which probably only is `ordiareatest`). Like you said, we also need synonymizing `plot = FALSE` and `draw = "none"` (with precedence to `plot = FALSE` if that is the future standard), and a deprecation period to drop `draw = "none"`. The `draw = "none"` indeed was just a quick hack for `ordiareatest`. However, there would be much less work in using `draw = "none"` in `ordicluster`. It is a matter of taste, and I accept either choice (much work or less work, if I don't have to work).

gavinsimpson:
OK. I'll work on that, harmonising to `plot = TRUE` but retain `draw = "none"` as an option that we mark as deprecated, and until it is defunct a `draw = "none"` will mean `plot = FALSE` regardless of what `plot` is set to (you'll get a warning any way).

jarioksa:
Should anything of this appear in **vegan_2.4-2**? I don't have a firm plan of ever having **vegan_2.4-2**, but we may have that before the major **2.5-0** release.

gavinsimpson:
@jarioksa it depends how long before you envisage 2.5-0 shipping? The bugfix to the current returned object should make it into 2.4-2. If I don't break backwards compatibility all of this could go into a 2.4-2. I don't envisage breaking anything with the changes proposed. If this goes in 2.4-2, we probably want to deprecate `draw = "none"` in 2.5-0 and make it defunct in 2.6-0 say.

Let me get this worked on and into the master branch and then we can chat about what gets ported to the 2.4-x branch.

jarioksa:
I am still a bit hesitant for adding a new argument (`plot`) when we already have an argument for the purpose (`draw`). In particular, if we want to extend this to other functions in the `ordi*` family, these would then have both `plot` (logical) and `draw` (text). The `plot = FALSE` argument is indeed used in some functions in **R**, but not in all. Think about `points(..., type = "n")` which certainly does not plot anything: just like `ordi*` functions, `points` is a function that **adds** items to an existing plot, and using `type = "n"` just means that it does nothing (except returns `invisible(NULL)`). For `boxplot` and `hist` `plot = FALSE`  suppresses setting up the plot, but `ordi*` functions never set up a plot: they need an existing plot to add items there.

jarioksa:
We don't have `draw` argument in `ordicluster` yet, because currently we only have one alternative to draw. Someday new choices may emerge, and then we should add `draw` there, too. Somebody could thing, for instance, that this is a fancy idea:
ordiclusterdrawellipse

jarioksa:
How do you get rid of this codecov crap?

gavinsimpson:
@jarioksa I added those checks to see how we do with coverage of the code base in our unit tests. I didn't realise that the default was to add a comment to a PR with a coverage diff. I believe I have now configured codecov to not add a comment to the PR thread with the coverage diff. I'll look into the codecov/patch and codecov/project checks to see if these can be turned off or not.

jarioksa:
@gavinsimpson : I detest codecov because it is based on the wrong concept -- testing in **R** packages is not done and should not be done like codecov assumes. codecov assumes that all tests are in the `tests` directory, and probably you also borrowed the failed **testthat** package assumption that tests should be simple statements returnin logical scalar (`TRUE`, `FALSE`). The **R** package concept is that the results of expressions are compared against reference output, and developer must see that there are no significant changes in the references output (there are round-off differences among platforms, for instance). The main tests are examples in `man/*.Rd` files which should show the main usage of the documented functions. There may be additional tests for boring corner cases in the `tests` directory, but it is only supplementary location for additional tests and not the main forum of tests. Further, also in `tests` directory the main mode of testing is seeing that output does not change. For instance,  `summary(eigenvals())` can fail in 36 ways and still it passes your new test (does not return a list and returns on object that inherits from `matrix`). A better test would have been
```R
data(varespec, varechem)
mod <- cca(varespec ~ Al + P + Condition(pH), varechem)
summary(eigenvals(mod))
```
which would alert developers if the eigenvalues or their summary changes in any way when compared against references output.

I think that the most damaging feature of codecov is that support a bad practice of producing ritualistic and useless tests -- tests for the sake of tests -- which only give a false feeling of safety and self-righteousness.

gavinsimpson:
Yep, I saw that but have been busy and hadn't gotten round to changing the test to print some output.

I don't fully agree with your assessment of tests; the examples section can't possibly contain all options or combinations of arts for even modestly complex functions. Furthermore, you can't test  that your code fails gracefully upon bad inputs in examples. I see a need for both types of checks/tests.

FYI, testthat can easily compare output if tests with reference output.

All codecov is doing is reporting how much of the codebase is actually exercised during checks. It is how one writes tests that determines the utility of those tests.

jarioksa:
I merged the `ordicluster` return object fix (00bf35199) to the cran-2.4 branch although this PR is not yet merged to the master.

jarioksa:
I'm starting to prepare the 2.5-0 release, and this PR should be resolved for that release. I am still not happy with the `plot = FALSE` argument, but I think that `draw = "n"` or `draw = "none"` is preferable. Please note that `ordicluster` **never** plots, but it **adds** or draws new items to existing plots. So having draw nothing makes sense.

gavinsimpson:
OK, let me see what I can do this week.

gavinsimpson:
I merged in the changes listed here with a modification from `plot = TRUE` to `draw = c("segments", "none")` as per discussion.

I'll make additional changes as needed to other functions, to have them return the information they prepare for plotting, in other commits/branches.

PR #222, opened 2017-01-13
Added iterations to rrarefy

Microbiology:
Some rarefaction tools allow the user to subsample their data sets using multiple random subsampling iterations and reporting the mean of the results.

I added this functionality to the `rrarefy` function. The user can now specify the use of more than one subsampling iteration to get the mean of the results. This is a relatively small update.

### DIFF
--- a/R/rrarefy.R
+++ b/R/rrarefy.R
@@ -1,32 +1,46 @@
 ### Random rarefied subsample: sample without replacement
 
 `rrarefy` <-
-    function(x, sample)
+    function(x, sample, iterations = 1)
 {
     if (!identical(all.equal(x, round(x)), TRUE)) 
         stop("function is meaningful only for integers (counts)")
     x <- as.matrix(x)
-    if (ncol(x) == 1)
+    if (ncol(x) == 1) 
         x <- t(x)
-    if (length(sample) > 1 && length(sample) != nrow(x))
-        stop(gettextf(
-             "length of 'sample' and number of rows of 'x' do not match"))
-    sample <- rep(sample, length=nrow(x))
+    if (length(sample) > 1 && length(sample) != nrow(x)) 
+        stop(gettextf("length of 'sample' and number of rows of 'x' do not match"))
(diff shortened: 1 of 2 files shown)


jarioksa:
This indeed seems to work and this is safe in the sense that it does not influence the default case `iterations = 1` . What I miss is a use case and motivation: why would somebody like to do something like this? Are there some applications where people do something like this or would like to do something like this?

There are some issues here which worry me. One of these is that `rrarefy` requires integer input, but this function will return non-integer result of integer data. `rrarefy` result is just a random sample from integer data, but what is this here? I know people in some fields misuse (like I see)  `rrarefy` to equalize sample sizes. I think this should not be done, but that solves no problem but brings some new problems. This function is helpful in the way that it reduces the random sampling error, but it in the same it loses the property of random sample. If you want to have equalized averages, why don't you calculate them directly: when `iterations` approach infinity, the result approaches `x/rowSums(x) * sample` or equivalently `decostand(x, "total") * sample`  (where `sample` is the subsample size argument of `rrarefy`). Why not go directly to the averages instead of their inaccurate estimation via `iterations`?

Microbiology:
That is a very good point, I have been over complicating things. It makes more sense to calculate the equalized averages as you described. I updated the pull request code to reflect the change.

As far as use case, I think [this is a good reference](https://www.mothur.org/wiki/Rarefaction.single).

Thanks for the input, I really appreciate it! :)

Microbiology:
Yeah you are definitely right and that simpler formula would be the best way to do it. I was definitely over thinking things. haha

So after thinking about it more, I think the current update might be a better way to do it. Subsample the averages and set everything below `(sample)^-1` as zero. That way the rarer species that would otherwise have been missed due to subsampling are excluded from the analysis.

Does this seem to be an okay approach?

Microbiology:
So I guess as an update on my end. I have been discussing this with more people and I think, like you were getting at above, this is not the best way to implement subsampling. I will work out how to do this correctly and get back to it.

Thanks for your input and consideration! I will close this request.

PR #230, opened 2017-03-03
Make tolerance.cca more robust to round-off errors

gavinsimpson:
Implement Jari's suggestions to make `tolerance.cca()` more robust to rounding errors.

### DIFF
--- a/R/tolerance.cca.R
+++ b/R/tolerance.cca.R
@@ -68,7 +68,7 @@ tolerance.cca <- function(x, choices = 1:2,
             y <- sweep(Y, 1, Ytot, "/")^2
             N2 <- 1 / rowSums(y, na.rm = TRUE) ## 1/H
             ## avoid almost-1 for sites with only one spp
-            N2 <- zapsmall(N2 - 1L) + 1L
+            N2 <- zapsmall(N2)
             ## avoid "negative zeros" form 1 - 1/N2 when N2 ~ 1
             res <- sweep(res, 1, sqrt(pmax(1 - 1/N2, 0)), "/")
         }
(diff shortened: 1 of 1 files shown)


jarioksa:
I don't think `zapsmall` will work: it is intended to zap small values when `print`ing a vector, but it does nothing to single small values which will be returned with their `digits` anyway:
```r
zapsmall(c(-.Machine$double.eps, runif(2)))
# [1] 0.0000000 0.6715691 0.4652509
zapsmall(-.Machine$double.eps)
# [1] -2.220446e-16
```
To zap small values you really need to zap them to zero.
```r
ZAP <- sqrt(.Machine$double.eps) # conventional wisdom
if(abs(x) < ZAP) x <- 0
```

R/tolerance.cca.R
```diff
@@ -84,11 +89,14 @@ tolerance.cca <- function(x, choices = 1:2,
         if(doN2) {
             y <- sweep(Y, 2, Ytot, "/")^2
             N2 <- 1 / colSums(y, na.rm = TRUE) # 1/H
+            ## avoid almost-1 for species present only once
+            N2 <- zapsmall(N2 - 1L) + 1L
```
jarioksa: N2 <- zapsmall(N2)

R/tolerance.cca.R
```diff
@@ -84,11 +89,14 @@ tolerance.cca <- function(x, choices = 1:2,
...
             ## avoid "negative zeros" form 1 - 1/N2 when N2 ~ 1
             res <- sweep(res, 1, sqrt(pmax(1 - 1/N2, 0)), "/")
         }
     }
     res[!is.finite(res)] <- 0 # some values can be Inf or NaN but are really 0
+    res <- zapsmall(res)      # almost-zero tolerances should be zero
```
jarioksa: this cannot be zapped because the non-zero bits are the first (and only) ones, and you need
res[res < sqrt(.Machine$double.eps)] <- 0 or similar.

R/tolerance.cca.R
```diff
@@ -66,16 +67,20 @@ tolerance.cca <- function(x, choices = 1:2,
         if(doN2) {
             y <- sweep(Y, 1, Ytot, "/")^2
             N2 <- 1 / rowSums(y, na.rm = TRUE) ## 1/H
-            res <- sweep(res, 1, sqrt(1 - (1/N2)), "/")
+            ## avoid almost-1 for sites with only one spp
+            N2 <- zapsmall(N2 - 1L) + 1L
```
jarioksa: This can be zapped to 1: zapping removes the last (non-significant) bits, but it cannot be zapped to 0 which keeps the first (and only) non-zero bits: So change to

N2 <- zapsmall(N2)

gavinsimpson:
@jarioksa's code review has been implemented as suggested.

I'll cherry pick this into the 2.4 branch once the above checks have run and I'm merged this with the devel (2.5) branch

gavinsimpson:
@jarioksa I've now cherry-picked the two commits into the cran-2.4 branch. If this was the last outstanding thing of note before releasing to CRAN I guess it is ready to be packaged for release?

PR #237, opened 2017-05-31
Added average distance matrix function.

Microbiology:
I love using the Vegan package but it was missing a functionality that I hoped to use. In the Mothur microbial ecology toolkit, the distance matrix calculation function incorporates random subsampling, such that it will create many distance matrix from many iterations of random subsampling, and then use the average of the matrices in a single new matrix. Here is a link to what I am talking about:

https://www.mothur.org/wiki/Dist.shared

I added a similar and simple function to vegan in this pull request. The function is called `avgdist` and will make running this type of analysis easier for other users (I hope). Hopefully you also find this useful.

FWIW, it passes when I run `devtools::check("./vegan")` with the added function. Thanks and I appreciate any feedback! :)


jarioksa:
The idea looks natural -- I know microbiologists do this. However, the current PR has several technical problems that must be solved:

- [ ] It adds 14 binary files by compiling C and Fortran source files. These shall not be in the source package. These files were generated because you ran `devtools::check("./vegan")` which checks the package within the source directory. What you must do is to first build the package into source package and check that: `cd .. ; R CMD build vegan; R CMD check vegan_2.5-0.tar.gz`. Alternatively, you should run `git clean -f` within the source directory to remove untracked files before you commit your changes.
- [ ] Similarly, `devtools::check()` adds a superfluous `RoxygenNote` line to `DESCRIPTION`. We do not use `roxygen` in **vegan** and this line should not appear in the `DESCRIPTION`.
- [ ] Indentation should be improved. Use offset of 4, and indent logically. Current function is difficult to read despite its brevity. Look at the indentation model from other **vegan** functions.
- [ ] Arguments in the function call should be arranged so that arguments without defaults are before those that have defaults. This makes life easier because you can rely on positional matching for compulsory arguments. I would also like to have default value to `iterations`, and for consistency with `rrarefy`, you may consider renaming `depth` to `sample` (but that is not necessary). So the sequence could be `avgdist(x, depth, iterations = 100, dmethod = "bray")`.
- [ ] I would like to have this function to work with *any* standard dissimilarity function, including those not in **vegan** but returning regular `"dist"` class objects and taking `method` as the second argument (in **vegan** this would mean that this works with `designdist`, `chaodist` and naturally in base `stats::dist`). So add an argument for selecting distance function alternative to default `vegdist`(see, e.g., `metaMDS`). This also means that you shall not list all distance methods in `man/avgdist.Rd` (which also is a maintenance nightmare because any changes in `vegdist` should be documented here also).

Microbiology:
Thanks so much for the feedback! This is really helpful, especially the info about devtools::check. I will work on making the changes.

Microbiology:
Hey jarioksa. Thanks again for the feedback! I updated the function and documentation to address your comments, and used your better checking method to avoid adding any newly compiled files.

The only thing I left out was the change from depth to sample, just because I thought it might get confusing in the code, since I call the rarefaction function and the resulting argument would be `sample = sample`.

Let me know how it looks to you and anything else I can do to improve it. :)

jarioksa:
I haven't tested this yet, but it looks basically OK. Here some quick comments:

- It is OK to use structures like `sample = sample` where the latter `sample` is an input argument.
- For safety, it is useful to have `distfun <- match.fun(distfun)`.
- I had `iterations = 100` just as an idea: how long does it take? Should the default be lower? If it takes long, it is useful to have lower number of iterations in the Examples of `man/avgdist.Rd` to cut down the package checking time.
- **vegan** has data sets `BCI` and `mite` with real count data: use one of these as an example instead of tweaking `varespec`.

Microbiology:
Hey jarioksa. Thanks for the comments! This has overall been a really helpful process so thanks! :)

I converted the `depth` parameter to `sample` and also added `distfun <- match.fun(distfun)` for safety. 

I think `iterations = 100` is a good default because it is robust enough for people to use. To speed things up in the examples though, I reduced it to ten.

I also added the `BCI` data, which definitely makes more sense.

Looks like it is passing tests on my end. Let me know what you think. Thanks again! :)

jarioksa:
This looks OK, except for one surprising feature: you silently remove sampling units that have lower total richness than your rarefaction depth? Why? After this, dissimilarities and the data do not match, and this will confuse users. Even experienced users can find it very difficult to find the reasons for this mismatch and resulting obscure error message in other analyses.

I really cannot understand the reason of removing sampling units (and I think we have discussed this very same issue earlier). The very minimum is to be loud about this removal, i.e., print a message. The result object should also include an attribute that tells which are the removed sampling units so that user really can use the result object later, and know how to match data and dissimilarities (there is a "Label" attribute which will contain the names of kept sampling units if these had names originally). I'd prefer just keeping those sampling units that were not rarefied. I find very little reason to any other action, but if there is such a reason, I'd like to learn the argument.

Microbiology:
Hey jarioksa, sorry for the delay is getting back to you. Our lab and others use the subsampling approach to bring all of the samples down to the same number of 16S rRNA gene sequences, for our bacterial ecology applications. For example, if we calculated diversity using subsampling at 10,000 and some samples only had 100 sequences, we would get skewed results and miss the rarer species. But I do also appreciate that this is an ongoing point of discussion, as discussed [here](http://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1003531).

I agree that this should at least be incorporated into the documentation and a warning should be printed. Also an option for doing this or not should also be added. I can work this into the code. :)

Thanks again for your feedback!

jarioksa:
I already merged this PR. My idea was to see these problems later, but if you wish, you can commit new changes to your PR and they can be merged later.

In your case the failed subsamples really are failed observations and should be removed. However, I do think that this data cleaning should be done with the data and not when you calculate a dissimilarity from the data. It is essential maintain the match between data and results used later with the data.

Currently `rrarefy()` function retains the sampling units that are smaller than the subsample, but it issues a warning. With `avgdist` it would issue a warning at every iteration which really is non-practical as you only get a message of "more than 50 warnings". The warning option should be turned off during iterations and warning should be issued only after iterations if some sampling unit was smaller than the subsample.

Issue #243, opened 2017-07-12
adding a square root transformation


sydneyg:
Hi @Microbiology  can you please add a capability to do a square root transformation of the median/mean of the subsampled OTU tables prior to making the Bray-Curtis distance Matrix?

Thanks for this function! It looks awesome! This is for the avgdist.Rd function

sydneyg:
Sorry this is the avgdist.Rd function

jarioksa:
The same comment as for issue #242: the `avgdist` function does not find median/mean of subsampled OTU tables, but mean of dissimilarities.

sydneyg:
Thank you Jari, that is correct. I want to find the median of the dissimilarity matrices rather than the mean of them.

Microbiology:
I added this functionality and included it in a pull request.

jarioksa:
Closed with PR #244

Issue #242, opened 2017-07-12
Switch to median instead of mean


sydneyg:
@Microbiology Hi can you please add an option to make a median of the subsampled OTU tables instead of the mean?

gavinsimpson:
Which functions did you have in mind for this?

sydneyg:
sorry, I've never written an issue before, this is the avgdist.Rd function.

gavinsimpson:
This seems trivial to add.

@Microbiology --- as you committed this function, do you have any input on whether this is a useful feature to add (& also #243)? (I'm not expecting you to implement it of course!)

jarioksa:
The `avgdist` function does not find mean of OTU table, but it calculates the dissimilarities for each subsampled OTU table and returns the average of these dissimilarities. In principle, it could return the median of these dissimilarities, but you were not suggesting this? (And this means having a new trick to calculate such a median).

jarioksa:
Replacing the line

```r
findist <- Reduce("+", distlist)/length(distlist)
```
with
```r
findist <- apply(do.call(rbind, distlist), 2, meanfun, ...)
```
and having new argument `meanfun = mean`  would allow user-selected function to find mean, and having those dots (`...`) would allow having mean functions that take extra arguments (such as `trim`).

Microbiology:
Thanks @sydneyg for the issue and the feedback (I had suggested this and the other issue be created in a previous conversation).

I will work on adding this in to the `avgdist` function, using the info from @jarioksa as well. :)

Microbiology:
I added this functionality and created a pull request to incorporate it (#244).

jarioksa:
Closed with PR #244

sydneyg:
Thank you for adding this feature to avgdist.R Microbiology! However, when I ran the script I got the same results regardless of using the mean or median options. Is this implemented correctly? Also, can you please put in the dcoumentation that Bray-Curtis is the default dissimilarity index?

Thanks! 

avgdist_7kmean <- avgdist(otu, 7000, meanfun= mean, transf= sqrt )
avgdist_7kmedian <- avgdist(otu, 7000, meanfun=median, transf= sqrt )

Microbiology:
Thanks for pointing this out @sydneyg ! :) I think you need to reopen the issue since you are the original poster.

I'll look into this and follow up.

Microbiology:
[...]

PR #244, opened 2017-07-29
Transformation and Median for `avgdist`

Microbiology:
I addressed issue #242 and #243 by adding an option to transform the count data and calculate median instead of mean. 😄 

The updates pass my R command checks.

### DIFF
--- a/R/avgdist.R
+++ b/R/avgdist.R
@@ -1,5 +1,5 @@
 `avgdist` <- 
-    function(x, sample, distfun = vegdist, meanfun = mean, iterations = 100, dmethod = "bray", ...)
+    function(x, sample, distfun = vegdist, meanfun = mean, transf = NULL, iterations = 100, dmethod = "bray", ...)
 {
     if (is.na(sample))
         stop("invalid subsampling depth")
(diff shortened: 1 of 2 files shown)


jarioksa:
Seems to work like intended. Some minor comments:

- it may be safe to use `meanfun <- match.fun(meanfun)` before using the `meanfun` function.
- the function takes now extra arguments (`...`), but these are ignored. You could use them in `output <- apply(afunc, 1:2, meanfun, ...)` to allow `meanfun` with extra arguments, such as `trim` in `mean()`.
- the use of `return` is not necessary and we do not have it in **vegan** unless necessary (*i.e.*, when exiting the function before reaching the end).

Microbiology:
Thanks for the feedback! :)

I made the suggested revisions and updated the pull request.

jarioksa:
Fine! Merging now. Thanks for your work!

jarioksa:
@Microbiology : I made some maintenance tweaks to `avgdist` and commited them to master as 8a0e3f004a9c88. Please check these.

Microbiology:
These look great, thanks @jarioksa! :)

Issue #246, opened 2017-08-13
avgdist silently drops observations


jarioksa:
Function `avgdist` will silently drop observations (rows) from the data. This can lead into mysterious mismatches with supplementary data sets. The minimum is to warn on dropping of data. The **vegan** standard is not to drop data, and the underlying `rrarefy` function does not drop observations. NB @Microbiology .

Microbiology:
Thanks for making this @jarioksa. Sorry, I know we had talked about it before but it fell through the cracks. I'll be able to take care of this sometime in the next couple of days.

jarioksa:
Closed with PR #251 (commit 2bcd09e6cdb6d10c).

PR #251, opened 2017-09-05
Addressing "avgdist silently drops observations"

Microbiology:
As described in issue #246, samples were being dropped silently if they failed to meet the subsampling threshold. I fixed that by including printed text identifying the removed samples.


R/avgdist.R
```diff
@@ -11,7 +11,10 @@
         transf <- match.fun(transf)
     # Get the list of iteration matrices
     distlist <- lapply(seq_len(iterations), function(i) {
-        inputcast <- rrarefy(x, sample = sample)
```
jarioksa: This fails examples: no `inputcast` found. You removed it from this line, and reference it later.

R/avgdist.R
```diff
@@ -23,10 +26,27 @@
...
     afunc <- array(
         unlist(as.matrix(distlist)),
                c(dim(as.matrix(distlist[[1]])), length(distlist)))
     output <- apply(afunc, 1:2, meanfun, ...)
+    # Set the names on the matrix
+    colnames(output) <- rownames(output) <- rnames
+    # Print any samples that were removed, if they were removed
+    if(nrow(x) != nrow(output)) {
+        dropsamples <- setdiff(row.names(inputcast), row.names(output))
+        write(
```
jarioksa: canonical way of printing warnings is to use `warning()`.

Microbiology:
Thanks for the comments. I went back and fixed the typos that were losing the data frames, and used `warnings` to print the warnings. Let me know how it looks, and thanks so much for your feedback! :)

Issue #259, opened 2017-12-01
rrarefy- fix silent failure with numeric count table


rrohwer:
There is a bug in the `rrarefy()` function if the structure of the count table is numeric/double instead of integer because of the behavior of the base r function `rep()` in line 18 of the function:  
```
row <- sample(rep(nm, times = x[i, ]), sample[i])
```  
rep always rounds down, so it gives unexpected results when a very large matrix of doubles that is basically whole numbers has tiny rounding errors. Here's a simple example of the unexpected behavior:  
```
> rep(x = letters[1:3], times = c(.99999999, 0.0000001, 1))
[1] "c"
```
This is a fix to get the desired behavior:
```
> rep(x = letters[1:3], times = as.integer(round(x = c(.99999999, 0.0000001, 1), digits = 0)))
[1] "a" "c"
```
This means that if you are subsampling to the shortest sample, rep will throw an error sometimes saying replace=F and there aren't enough things to sample from. I suggest fixing it by adding this after the other checks in the `rrarefy` function:
```
if (as.numeric(x)){
   x <- round(x, digits = 0)
   save.row.names <- rownames(x)
   x <- apply(X = x, MARGIN = 2, FUN = as.integer)
   rownames(x) <- save.row.names
   rm(save.row.names)
}
```

Thanks!
Robin

jarioksa:
An intriguing point of view. Can you trigger this error with real data? You mean this is not sufficient to catch non-integer data:
```r
    if (!identical(all.equal(x, round(x)), TRUE)) 
        stop("function is meaningful only for integers (counts)")
```

gavinsimpson:
...and given @rrohwer's area of research I can well imagine such issues being all too common an occurrence in real data.

rrohwer:
Yes it occurred with my data. It happened because I had normalized the data instead of rarefying, but later converted the normalized OTU table back to counts to simulate how results changed with different sequencing depths. Since the data table was very large, I was getting a difference of over 30,000 between the length and the sum!! 

Since the first check:
```
if (!identical(all.equal(x, round(x)), TRUE)) 
    stop("function is meaningful only for integers (counts)")
```
makes sure you are using whole numbers already, I think it would be OK to convert to integer structure automatically. Plus, I think most downstream things that might requiring a numeric structure would automatically convert it back.

jarioksa:
Good morning, America!

I think this is the simplest fix:
```diff
[...]

PR #260, opened 2017-12-03
Issue #259

jarioksa:
This PR implements soft testing of integer input, and if passed, hard fix to integers with `round()` like I outlined in PR #259. The fix is implemented for `rrarefy`, `as.preston` and `estimateR`  where exact integers were needed because of truncation or breaking ties exactly at the border. This also uses slacker test for `vegdist` methods that use integers, because older truncation invariance was not needed. Other methods can cope with nearly-integer data.


gavinsimpson:
This looks fine to me.

PR #286, opened 2018-07-24
decostand method "freq" does not exist in code

KasperSkytte:
Either adjust the code as in this PR to match documentation or the opposite. The documentation of `decostand` says "freq" and not "frequency"

### DIFF
--- a/R/decostand.R
+++ b/R/decostand.R
@@ -3,13 +3,13 @@
 {
     wasDataFrame <- is.data.frame(x)
     x <- as.matrix(x)
-    METHODS <- c("total", "max", "frequency", "normalize", "range", "rank",
+    METHODS <- c("total", "max", "freq", "normalize", "range", "rank",
                  "rrank", "standardize", "pa", "chi.square", "hellinger",
                  "log")
     method <- match.arg(method, METHODS)
     if (any(x < 0, na.rm = na.rm)) {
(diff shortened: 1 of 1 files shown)


jarioksa:
`"freq"` may not *exist*, but it *works*:
```r
data(dune)
fdune <- decostand(dune, "freq") # OK
```
After you fix, this would fail:
```r
fdune <- decostand(dune, "frequency") # works now, fails after this PR
```
What about fixing the documentation instead?

KasperSkytte:
`match.arg` catches it anyways because it only performs partial matching, I just couldn't find `"freq"` anywhere in the code and thought it's best to be explicit. Perhaps it's better to adjust the documentation then. Or leave it, it does work as intended, up to you.

jarioksa:
The risk of regression is really small in this case. However, if we change the documentation to  `"frequency"`, it will work both with `"freq"` and `"frequ"` or longer, but if we change the code to `"freq"`, it will fail with `"frequ"`. It may be that this failure is an empty set in scripts and applications around, but changing documentation is still safer than changing code.

KasperSkytte:
I agree. Changing documentation is the safer choice. But users must have read the documentation before they would know that the method is available in the first place. So hopefully everyone uses "freq", but we can't be sure. I made a new PR.

jarioksa:
see PR #287 (merged)

PR #294, opened 2018-11-05
SIMPER: new implementation

jarioksa:
The simper() function was completely re-written to make permutation tests faster. They indeed are an order of magnitude faster, and permutations are now run as default.

The current function can also analyse non-grouped data. These show the species contribution to the average dissimilarity (overall heterogeneity) in the data. Grouped data only studies dissimilarities among sampling units in different groups, and for several species these are higher. This can be seen by comparing contributions with grouped and non-grouped data. In non-grouped data the contribution is related to within-species heterogeneity (variance), and is usually highest for the most abundant species.

The documentation warns now more clearly against using simper. The danger of simper is that people may take it as finding species that differentiate groups. This is not done in simper, but it reports the contributions to total dissimilarities where the major component usually is the within-species variance irrespective of any class structure. I guess that most users have always misunderstood the results and reached wrong conclusions.

@EDiLD have a look at this if you only have time.


jarioksa:
The failed test on Travis is not a vegan problem, but it is Travis problem: getting gpg fails and R and cannot be fetched and build.

PR #298, opened 2018-12-21
add color of ordilabel and legend in plot.betadisper function.

zhilongjia:
as show in Figure. This is helpful if there are several groups. Thank you. 
[https://i.ibb.co/f1Zq3vp/Screenshot-from-2018-12-21-22-16-16.jpg
image
](url)

### DIFF
--- a/R/plot.betadisper.R
+++ b/R/plot.betadisper.R
@@ -94,8 +94,12 @@
     }
     points(g$sites, pch = pch[x$group], cex = cex, col = col[x$group], ...)
     if (label) {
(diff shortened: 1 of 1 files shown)


gavinsimpson:
Thanks for the suggestion and patch.

I agree with changing the label colour, but I think the legend should be done by the userif needed as it provides no further information than the centroid label.

Do you want to modify your patch to include just the label colouring bit?

zhilongjia:
Ok, see #300, Thank you.

Issue #299, opened 2018-12-28
R: new formula & vegan breaks


jarioksa:
I just got the following email message from Martin Maechler:

> just a quick note before I embark on longish trip to Calgary Canada:
Following up the thread on the R-devel mailing list,
I've recently changed  formula(<model.frame>) behavior such that it
returns the formula from the "terms" attribute of the model frame,
rather than a newly constructed formula from the model.frame's
variables (as it does for non-model.frame  data frames).

> You may have noticed on the CRAN checks for vegan that this breaks at
least one of vegan's examples on  "R-devel" (R's development version).

> I may take time on the long plane flight to see how to fix it... or
may not:   vegan seems to be doing pretty "strange" contortions in
evaluating formulas,
as far as I could tell, quite a bit non-standard and different than
say lm(), glm() etc.

This may mean that we have to do something with the formula. Most likely this concerns `cca.formula` and its ilk. Other functions have more usual handling of `formula`, although we do have matrix-like right-hand-side in most functions.

gavinsimpson:
The proximal issue of the error is in `ordiParseFormula()` on [L76](https://github.com/vegandevs/vegan/blob/master/R/ordiParseFormula.R#L76). There we call
```r
mf <- model.frame(formula(mf), mf, xlev = xlev,
                  na.action = na.action, drop.unused.levels = TRUE)
```
and in R-patched `formula(mf)` returns
```r
> formula(mf)
Management ~ `poly(A1, 2)` + spno
```
whereas in R-devel it returns
```r
> formula(mf)
~Management + poly(A1, 2) + spno
```

The current behaviour seems wrong &mdash; why is `Management` now the response? &mdash; but the direct issue is that currently we expect `poly(A1, 2)` to be quoted as we want it to find the variable (a matrix) of that same name in `mf`:
```r
> str(mf)
'data.frame':   13 obs. of  3 variables:
 $ Management : Factor w/ 4 levels "BF","HF","NM",..: 1 4 4 NA 2 2 1 4 4 3 ...
 $ poly(A1, 2): num [1:13, 1:2] -0.1421 -0.0579 -0.0684 0.1526 -0.0579 ...
  ..- attr(*, "dimnames")=List of 2
  .. ..$ : NULL
  .. ..$ : chr  "1" "2"
....
```
instead of re-evaluating the formula &mdash; we don't store `A1` or at least it's not available by the time we evaluate the model frame on L76. If it was R-devel would work.

[...]

PR #300, opened 2018-12-29
Color ordilabel in plot.betadisper function

zhilongjia:
color the ordilabel figured below.
image
Updated version for #298


jarioksa:
Basically OK. I just wonder what should we do with `ordispider` & `ordihull`, @gavinsimpson? They are no inconsistent with this one. The `col` argument here concerns `ordispider` label, but not its lines, and it also concerns `ordihull` border (but the function does not use `ordihull` labelling which is different from the centroid of points).

jarioksa:
Travis error seems to be due to issue #299 and not related to this PR.

PR #325, opened 2019-08-27
Parallel version of cascadeKM

becarioprecario:
Hi,

I have just modified the cascadeKM function to support parallel computing. It should work both on Unix (via mclapply) and Windows (via parLapply). 

Best,

Virgilio

### DIFF
--- a/R/cascadeKM.R
+++ b/R/cascadeKM.R
@@ -1,5 +1,6 @@
 "cascadeKM" <-
-function(data, inf.gr, sup.gr, iter = 100, criterion="calinski")
+function(data, inf.gr, sup.gr, iter = 100, criterion="calinski",
+  parallel = getOption("mc.cores"))
 {
 ### DESCRIPTION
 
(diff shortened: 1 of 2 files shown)


jarioksa:
Thanks for this work. However, its user interface does not accord to **vegan** conventions –  see `vignette("decision-vegan", package="vegan")`. What do you think about this? Would it be feasible to implement with our conventions?

becarioprecario:
Thanks for this work. However, its user interface does not accord to vegan conventions – see vignette("decision-vegan", package="vegan"). What do you think about this? Would it be feasible to implement with our conventions?

Yes. I could add a ‘parallel = getOption("mc.cores")’ argument to the function and then handle parallel computing for unix and windows depending on its value. Will that be fine?

becarioprecario:
Please, could you check that my latest modifications comply with your conventions?

becarioprecario:
@jarioksaplease, could you check that my latest changes follow the **vegan** convention and that it can be merged into the main repository? If not, I can make any required modifications.

jarioksa:
The implementation does not conform with the standard as described in the vignette on design-decisions, but I'll fix that after this merge. For instance, the numeric value of `parallel` will be ignored in unix-like systems, and user-supplied `makeCluster()` cluster will be ignored in unix-like systems and destroyed after use in Windows.

jarioksa:
@becarioprecario I added your name in the Authors filed in man/cascadeKM.Rd. That is OK? And also check the form I used.

becarioprecario:
@jarioksa yes, many thanks for checking and fixing my pull request and for adding my name to the Authors (maybe use V. Gomez-Rubio, orchid 0000-0002-4791-3072 if needed).

PR #350, opened 2020-03-25
Add names and clean rarecurve objects

fkeck:
PR to improve the object returned (invisibly) by the rarecurve function. It can be useful for the user to capture it to make custom plot (eg. use ggplot2).

Proposed changes:
- In each element of the list, get rid of the name attribute for Subsample which is meaningless.
- keep track of the site names by setting names to the returned list (i.e. rownames of x).

### DIFF
--- a/R/rarecurve.R
+++ b/R/rarecurve.R
@@ -30,8 +30,10 @@
         n <- seq(1, tot[i], by = step)
         if (n[length(n)] != tot[i])
             n <- c(n, tot[i])
+        attributes(n) <- NULL
         drop(rarefy(x[i,], n))
     })
+    names(out) <- rownames(x)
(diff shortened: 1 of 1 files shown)


gavinsimpson:
Sorry for the radio silence (I guess the pandemic was just kicking off and I never circled back to this PR being occupied elsewhere) --- I actually fixed (in a different way) the first part of this PR a few days after you posted it, via 7180c4e8929df1eab98ee637b5abd37678d525af. And @jarioksa has since made it much easier to grab a tidy version of the rarecurve information via the implementation of a `tidy` argument. So as I see it, there's no need now for the second part of this PR; is that right?

We are planning to release the current GitHub version to CRAN imminently so I will close this. Feel free to reopen or comment further if I'm missing something.

fkeck:
Thanks, the tidy argument is really useful.

Issue #433, opened 2021-08-14
Aitchison and Robust Aitchison distance?


antagomir:
Aitchison distance and its robust version have become frequent choices in the analysis of microbial communities, where compositional data is ubiquitous. The ability get access to these through a dedicated existing package would be a better solution than creating new implementations, and allow seamless linking with other packages that rely on `vegan::vegdist`.
 
Aitchison distance needs pseudocount in many applications. An alternative, "robust Aitchison distance" has been proposed in the literature, the difference is that CLR transformation is done only on the non-zero values. This has gained some attention recently, see e.g. [Martino et al. (2019)](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6372836/) and from there links to more original references on robust CLR / robust Aitchison.

Would you consider adding Aitchison and Robust Aitchison distance as new options in `vegan::vegdist` ?

jarioksa:
Quick look at the paper indicates that the index can be calculated as a Euclidean distance after "CLR transformation". In **vegan** design this potentially means implementing CLR transformation in `decostand`. This route would be used if CLR transformation is regarded as useful outside distance calculations. Are you considering to submit the code as a pull request?

antagomir:
Yes, Aitchison distance equals to Euclidean distances in CLR-transformed data.

Yes, CLR transformation is useful also as such, outside of distance calculations. Despite its limitations (requirement of pseudocount etc) it is frequently used at least in microbial community analysis to remove compositionality bias in relative data and enhance statistical comparisons between samples. Also other log-ratio transformatins (ALR, ILR, phILR) are sometimes used for this.

But there are already several R packages that provide CLR and other log-ratio transformations. One of them is **compositions**. If you would prefer avoiding new dependencies, we could first check if this happens to be included in any of the existing dependencies. But these transformations are simple and should be relatively straightforward to implement directly in **vegan**, if necessary. 

We can consider submitting the code as PR if this suggestion finds support.

jarioksa:
A comment about adding dependency: **compositions** adds a huge number of *chained* dependencies (*i.e.* it depends on packages that depend on packages that depend on packages that ... break). The transformations should be added independently or using a more light-weight dependency. The greatest complications on those transformations seem to be handling log(0) and 0/0.

gavinsimpson:
I'm pretty sure people doing this in community ecology just solve the log(x) for x=0 problem with the usual continuity correction and run everything with (equivalent of) log1p(x).

[...]

PR #458, opened 2022-01-07
Aitchison distance, clr transformation, and their robust variants

antagomir:
This PR solves #433 and provides:

New transformation to `vegan::decostand`:
- CLR transformation ("`clr`")
- robust CLR transformation ("`rclr`") 

New distance to `vegan::vegdist`:
- Aitchison distance ("`aitchison`")
- robust Aitchison distance ("`aitchison_robust`")

These options are documented, references and unit tests have been provided. The automated build & checks seem to go through.

Remaining questions:
- Are contributors listed in DESCRIPTION (I could add this to the PR)? The newer formats often include authors with different roles in this file (see [here](https://www.oreilly.com/library/view/r-packages/9781491910580/ch04.html)), like "creator", "author", "contributor" etc. But the current system does not use this format so I am not sure. Can also add an update to the full author listing format, if this is desirable and could fit in the same PR.
- NEWS file update? Would require opening up a new section with an updated version number.
- Any other issues, happy to fix!

Also thanks to the colleagues Felix Ernst & Tuomas Borman for contributions to the code via the *mia* package, from which these were transferred and modified.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -15,3 +15,4 @@ Description: Ordination methods, diversity analysis and other
 License: GPL-2
 BugReports: https://github.com/vegandevs/vegan/issues
 URL: https://github.com/vegandevs/vegan
+RoxygenNote: 7.1.2
--- a/R/decostand.R
+++ b/R/decostand.R
@@ -89,8 +89,8 @@
         if (missing(MARGIN))
 	    MARGIN <- 1
         if (MARGIN == 1) 
-            x <- t(.calc_clr(t(x)))
-	else x <- .calc_clr(x)
+            x <- t(.calc_clr(t(x), ...))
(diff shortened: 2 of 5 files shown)


jarioksa:
This looks OK for a merge. There is one bug, but I'll fix that after merge (`vegdist.R` needs `{}` after `if()` when you have more than one statement – R is not python where indentation defines the block structure).

As to authors in `DESCRIPTION`: I'll re-write it to use `Authors@R` and then when we can update author information more easily. We have discussed this with @gavinsimpson , but we have just been lazy for the change. Wait for that.

Thanks for the contribution!

antagomir:
Cool, I added curly brackets now, I hope I didn't miss any. Looking fwd to seeing this in CRAN and taking into use in other pkgs - !

PR #464, opened 2022-01-28
Run R CMD check on MacOS X, Ubuntu, & Windows using GH Actions

gavinsimpson:
Appveyor seems to be very fickle at the moment, failing for obscure reasons as often as it runs without error. Travis is no longer the flavour of the month CI-wise.

This PR configures the main vegan repo to run R CMD check on the vegan sources using a GitHub Actions workflow. It checks the package under 5 systems:

* MacOS X (latest) with r-release
* Windows (latest) with r-release
* Ubuntu (latest LTS, currently focal IIRC)
    * r-release,
    * r-devel
    * r-oldrel-1 (the last minor release; if r-release is 4.1.x then this is R 4.0.x I believe)

If you have no problem with this @jarioksa I'll deactivate Appveyor and Travis from running checks.

I'm using these r-libs (RStudio) provided actions so I'm pretty familiar with how they work and they're pretty stable now (we're using the v2 release)


jarioksa:
Thanks! I was thinking about the same, and actually had a look at the issue just yesterday, but didn't make it, but got an error message when I tried to push the changes (GITHUB_TOKEN did not allow changes in `workflows`). I was doing everything from bash which is the environment I work in (plus emacs). I'll have peak quickly, and tell you how I feel about this.

jarioksa:
This looks fine to me, but needs some steps more: appveyor & travis should be disabled, and the check badges in README.md should be replaced with the gh action badge.

gavinsimpson:
Thanks @jarioksa I'll do those steps (I'll include fixing the README in this PR) and then merge

PR #493, opened 2022-03-01
Fix #492 (remove ... and add q and r as formal arguments to internal function getF)

### LINKED ISSUE #492: getF internal function should be passed q and r
`getF()`, internal to `permutest.cca`: https://github.com/vegandevs/vegan/blob/710877a1d401b978f15f0cbcfbfad23accc7e8a1/R/permutest.cca.R#L37-L55

has argument `...` which is not used, while `q` and `r` (the vector of rank - 1 degrees of freedom for each term, and the rank of the residual term - 1 respectively (?)) are *not* passed to the function. Instead, `getF()` relies on finding the values from the calling environment.

Propose this internal function is changed to pass `q` and `r` via new formal arguments to `getF()` and remove the formal `...` argument to `getF()`

gavinsimpson:
This commit fixes #492 

* removes `...` as a formal argument to `getF()`
* adds formal arguments `q` and `r` to `getF()`
* updates all calls to `getF()` in `permutest.cca` to pass `q` and `r` to these new arguments, including in all anonymous wrappers
* fixes one instance of `1:nperm` -> `seq_len(nperm)` for safety

I have run all checks and there are no changes in output following the changes in this patch.

@jarioksa would you mind giving this a once over as the internals of the permutation testing code for these models has changed an awful lot since I last looked at them?

### DIFF
--- a/R/permutest.cca.R
+++ b/R/permutest.cca.R
@@ -34,8 +34,10 @@ permutest.default <- function(x, ...)
     isPartial <- !is.null(x$pCCA)   # handle conditions
     isDB <- inherits(x, c("dbrda")) # only dbrda is distance-based
     ## C function to get the statististics in one loop
-    getF <- function(indx, ...)
+    getF <- function(indx, q, r)
     {
+        # q is the rank(s) of the effect(s) - 1
+        # r is the rank of the residual term - 1
         if (!is.matrix(indx))
             indx <- matrix(indx, nrow=1)
         out <- .Call(do_getF, indx, E, Q, QZ, effects, first, isPartial, isDB)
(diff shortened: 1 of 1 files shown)


jarioksa:
This proposes a fix for something that works. Moreover, the fix is only partial: the internal function also needs `E`, `Q`,  `QZ`, `effects`, `first`, `isPartial`, `isDB`  that are arguments of `.Call(do_getF,...)`, but are not in the `getF` call.

The dots may not needed, though, but I haven't checked all instances where `permutest.cca` is called and may pass arguments to `getF`.

gavinsimpson:
Gah, yes I missed those argument - had a blind spot for the C code call. I'll fix that.

I'll double check the `...`; I couldn't see any issues when I looked but I should double check....

The reason I'm proposing to fix this is really stylistic; R has lexical scope that allows this, but from a functional programming view point, we should be passing in all arguments needed by a function so that it's clear what is need and where it is coming from.

gavinsimpson:
Also, I see to have merged this (I pushed something else to upstream and had forgotten I'd merged this locally). I'll fix that *sigh*

jarioksa:
Check parallel processing in Windows, if only possible (`makeCluster` can be quirky with parameters). I guess it *may* be sufficient to test on Linux with
```r
p <- makeCluster(2) # socket cluster: the only one that works in Windows
permutest(<mod>, parallel = p) # use socket cluster 
```

PR #460, opened 2022-01-15
Aitchison

antagomir:
This solves the naming issue in #433 

Method names in `vegdist` are now changed from "aitchison" and "aitchison_robust" to "Aitchison" and "rAitchison", respectively.

In addition, I added the transformations "alr" and "ilr" due to recent requests. These were discussed in the original issue as well.

In addition, I updated unit tests for all these methods. The tests include also comparisons with alternative packages, to show the compatibility (those tests with external packages are commented out, though). 

Everything seems to work and properly tested, but cross-check always good.

PR #498, opened 2022-03-02
Aitchison

antagomir:
Summary of this PR:

1) Switches to the suggested naming in [#460](https://github.com/vegandevs/vegan/pull/460) ie. "aitchison" and "robust.aitchison"

2) Removes ILR related parts that were in my earlier PR as the implementation was also simple and unoptimized. Better to add ILR more properly if a need will arise.

3) Updated tests and documentation accordingly w.r.t. 1-2

I think this PR should resolve all the remaining issues that there were.

### DIFF
--- a/R/decostand.R
+++ b/R/decostand.R
@@ -5,12 +5,12 @@
     x <- as.matrix(x)
     METHODS <- c("total", "max", "frequency", "normalize", "range", "rank",
                  "rrank", "standardize", "pa", "chi.square", "hellinger",
-                 "log", "clr", "rclr", "alr", "ilr")
+                 "log", "clr", "rclr", "alr")
     method <- match.arg(method, METHODS)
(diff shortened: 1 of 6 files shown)


antagomir:
This PR adds to #460

gavinsimpson:
Just following up on this; does this need merging in addition to #460 ? The comment by @antagomir suggests it is but I lost track of these things.

antagomir:
Yes it does. This includes changes that were requested after merging #460

gavinsimpson:
Ah, OK, apologies for not noticing that. We'll take a look at get this merged ASAP (it doesn't appear to be causing any problems with our checks)

gavinsimpson:
Everything looked fine; merged. Thanks @antagomir for these contributions and for bearing with us getting the "t"s crossed and the "i"s dotted.

PR #500, opened 2022-03-09
Use !anyNA() over all(!is.na(.))

MichaelChirico:
`all(!(.))` is logically equivalent to `!any(.)`; in this case, that means we can use the specialized `anyNA()` instead as well

### DIFF
--- a/R/plot.cca.R
+++ b/R/plot.cca.R
@@ -117,7 +117,7 @@
         biplabs <- ordiArrowTextXY(mul * g$regression, rownames(g$regression))
         text(biplabs, rownames(g$regression), col = rcol)
     }
-    if (!is.null(g$centroids) && all(!is.na(g$centroids)) && type !=
+    if (!is.null(g$centroids) && !anyNA(g$centroids) && type !=
         "none") {
         if (type == "text")
             text(g$centroids, rownames(g$centroids), col = "blue")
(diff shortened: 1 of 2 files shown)


gavinsimpson:
There's also an instance in `R/ordConstrained.R` https://github.com/vegandevs/vegan/blob/48b7543121aec59e61dd65607ad76cb4d7579b86/R/ordConstrained.R#L248

Does this make a practical difference? - Are there examples of this being a performance bottleneck?

MichaelChirico:
to me it's about readability as much as performance -- I find "aggregate + negate" easier to process than "negate + aggregate".

Came across it because our (very) old version was tripping the previous error from `if (!is.na(wa))` when `length(wa)>1`, not for performance reasons.

gavinsimpson:
OK - I'll deal with this and the extra instance I found (plus a more thorough check for other instances than the GH search I did to find the `ordConstrained` one).

Thanks :-)

MichaelChirico:
FWIW this will eventually be available as the `OuterNegationLinter`:

https://github.com/r-lib/lintr/issues/884

the problem with writing all those linters is now I can't help but see them everywhere :)

gavinsimpson:
Thanks @MichaelChirico

jarioksa:
Just a historic note: `anyNA` was introduced in **R** version 3.1.0 in April 2014, and the code you changed is from 2001. I think this is "syntactic sugar", although the docs say that

> The default method for `anyNA` handles atomic vectors without a
     class and `NULL`.  It calls `any(is.na(x))` on objects with
     classes and for `recursive = FALSE`, on lists and pairlists.

It is more readable, though.

gavinsimpson:
Yep; I triggered the GH Actions checks for the PR to insure that we weren't inadvertently bumping the minimum version of R we require.

Issue #528, opened 2022-09-20
simper function returning an error using example dataset


camillemellin:
Hello,
I used to be able to successfully run this function, which now returns an error using the example dataset:

```r
>data(dune)
>data(dune.env)
>(sim <- with(dune.env, simper(dune, Management, permutations = 99)))

Error in array(r, dim = d, dimnames = if (!(is.null(n1 <- names(x[[1L]])) & : length of 'dimnames' [1] not equal to array extent
```

Am I missing something, or is this error due to a package update?
Many thanks for your assistance with this.
CM

jarioksa:
I cannot replicate this, but copy paster of your message runs smoothly without error. Need more info.

EDiLD:
I can reproduce this with the current CRAN version

```
R> example("simper", "vegan")
Loading required package: permute

Attaching package: ‘permute’

The following object is masked from ‘package:devtools’:

    check

Loading required package: lattice
This is vegan 2.6-2

simperR> data(dune)

simperR> data(dune.env)

simperR> (sim <- with(dune.env, simper(dune, Management, permutations = 99)))
Error in array(r, dim = d, dimnames = if (!(is.null(n1 <- names(x[[1L]])) &  : 
  length of 'dimnames' [1] not equal to array extent

R> packageVersion("vegan")
[1] ‘2.6.2’

R> sessionInfo()
R version 4.2.1 (2022-06-23)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 20.04.4 LTS

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/openblas-pthread/libblas.so.3
LAPACK: /usr/lib/x86_64-linux-gnu/openblas-pthread/liblapack.so.3

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8     LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                  LC_ADDRESS=C               LC_TELEPHONE=C             LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] vegan_2.6-2     lattice_0.20-45 permute_0.9-7 
[snip]
[...]

PR #530, opened 2022-09-27
Fix for proxy package breaking `simper()`

gavinsimpson:
This PR implements a simple fix to `simper()` that is similar to that suggested by @EDiLD in #528 but using the "guidance" of R Core to use `as.vector()` when one wants the side effects of `c` `?c`.


jarioksa (approved):
Looks good. Indeed, `c()` can have method functions which do not behave like assumed – somebody could make `c.dist()` – and hence `as.vector` is safer.

PR #571, opened 2023-04-04
Use ISHFT for powers of 2

MichaelChirico:
Context: a bug in our build system led to vegan failing to compile in some cases because of this exponentiation.

While that bug is our own to fix, it does strike me that `ISHFT()` is preferable for calculating powers of 2 in particular (and it solved our issue to boot):

https://gcc.gnu.org/onlinedocs/gfortran/ISHFT.html

### DIFF
--- a/src/monoMDS.f
+++ b/src/monoMDS.f
@@ -382,7 +382,7 @@ C
       IF (N.LT.2) RETURN
       FN=REAL(N)
       NLOOPS=MAX(NINT(LOG(FN)/LOG(2.)),1)
-      M=ISHFT(2,NLOOPS-1)
+      M=ISHFT(1,NLOOPS-1)
       DO II=1,NLOOPS
         FM=M
         DO I=1,MAX(1,N-M)
--- a/src/monoMDS.f
+++ b/src/monoMDS.f
@@ -382,7 +382,7 @@ C
       IF (N.LT.2) RETURN
       FN=REAL(N)
       NLOOPS=MAX(NINT(LOG(FN)/LOG(2.)),1)
-      M=ISHFT(1,NLOOPS-1)
+      M=ISHFT(1,NLOOPS-1) ! i.e., 2^(N_LOOPS-1) using bit shifts
       DO II=1,NLOOPS
         FM=M
         DO I=1,MAX(1,N-M)
--- a/src/monoMDS.f
+++ b/src/monoMDS.f
@@ -382,7 +382,7 @@ C
       IF (N.LT.2) RETURN
       FN=REAL(N)
       NLOOPS=MAX(NINT(LOG(FN)/LOG(2.)),1)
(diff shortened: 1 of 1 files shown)


gavinsimpson:
These check errors are unrelated, at least as far as I can tell - I'll update the workflows to the latest versions to see if that fixes things

jarioksa:
Looks OK, and the only thing that we lose is readability: ISHIFT(1, NLOOPS-1) is much more cryptic than 2**(NLOOPS-1) – but there is no great need for this specific subroutine be readable. I think it is surprising that the original code fails in your build system. As you notifeid, there was a problem in gfortran, but was tagged SOLVED in 2013. Fortran can use different internal code in A**B when A and B are both integer, and when they are real (double), and the compiler should be able to correctly choose the branch used in analysis. If it does not do the right thing, this is either a bug in the compiler or you have a broken build system. However, the easiest option is for us to adapt and switch to ISHIFT.

jarioksa:
Looks OK, and the only thing that we lose is readability: ISHIFT(1, NLOOPS-1) is much more cryptic than 2**(NLOOPS-1) – but there is no great need for this specific subroutine be readable. I think it is surprising that the original code fails in your build system. As you notifeid, there was a problem in gfortran, but was tagged SOLVED in 2013. Fortran can use different internal code in A**B when A and B are both integer, and when they are real (double), and the compiler should be able to correctly choose the branch used in analysis. If it does not do the right thing, this is either a bug in the compiler or you have a broken build system. However, the easiest option is for us to adapt and switch to ISHIFT.

jarioksa:
For every person reporting a problem, there are a thousand of people who have the same problem, but never report it. We'll merge this fix.

In Fortran `A**B` for `REAL` (`DOUBLE PRECISION`) arguments may expand to `exp(log(A) * B)`, but for integer exponents it can expand to a series of multiplications, and seems to do so in this case judging from function name ending `_i4_i4`. However, there may be an upper limit to an integer exponent (there used to be when I started programming in Fortran decades ago), but all cases should be handled smoothly. However, handling `2**INTEGER` is still a special case, and indeed handled most economically with `ISHIFT`.

jarioksa:
@gavinsimpson I know it is the Right Thing (TM) to use CI tools, but I have a feeling that we are rather testing the integrity of CI tools than **vegan**. All github error reports for several years have been caused by changes or breakages in fragile CI tools. I do test **vegan** after every commit I make locally and also after every pull from github – though only in one system – and I can see that the reported failures are failures in fragile and volatile CI tools. I think that now we should at least switch from `@v2`  tools to `@v3`, but I don't care to fix something I don't trust. Perhaps you can do this, and certainly you know these tools better than I do and can fix the other problems of the CI tools. Moreover, there really is no reason to test with **R** 3.5 or 3.6, nor with any explicitly numbered **R** release.

Issue #570, opened 2023-04-02
Adding matrix completion in rclr/rAitchison


antagomir:
The issue #433 has received request (from @johannesbjork) to add robust matrix completion in the robust CLR (rclr) method in the _vegan::decostand_ function, following the procedure in [Martino et al. (2019)[https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6372836/]. I agree with this suggestion.

The R package [ROptSpace](https://cran.r-project.org/web/packages/ROptSpace/index.html) includes an implementation of the OptSpace robust matrix completion algorithm used by Martino et al. (2019). There are many variants of matrix completion for missing values but this solution has been recently promoted in microbial ecology, and I am not immediately aware of other solutions that should be prioritized over this one.

It would be easy to add matrix completion in the rclr transformation in _vegan::decostand_ by using the ROptSpace functionality. However, this would introduce a new dependency in _vegan_, and this may not be ideal.

Fortunately, it seems that the _ROptSpace_ R package has been released with the MIT license and _vegan_ with GPL-2. [License compatibility](https://pitt.libguides.com/openlicensing/MIT#:~:text=The%20MIT%20License%20is%20highly,project%20must%20of%20GPL%20compliant.) allows including MIT licensed code in GPL-2. It seems possible to add this function in _vegan_ and cite the original source, then use this to add the matrix completion step in _decostand::rclr_ .

I can implement this change and open a PR but would like to cross-check first if the vegan developers?

handibles:
Hey Devs & co.,

Thanks as ever for the hard work. Came across this issue today (#433, #570) - as stands `decostand(x, "rclr")` replaces the 0 values with the *average* abundance within that sample ([line 191](https://github.com/vegandevs/vegan/blob/003391b92dc9d8ea20038865698f5e2fec95388b/R/decostand.R#L191)), as matrix completion is not implemented.

This seems fairly dangerous given that there's no warning in the output or documentation - the average vegan user, being trusting and kind, would have no reason to suspect their data was suddenly a strange, non-sparse beast.

Could a note be added in the meantime to flag this?... I could mock up a PR or similar, but not sure warranted.

antagomir:
Thanks @handibles -  I agree, and planning to go through this during the Xmas vacations if the situation allows. If there are further bottlenecks we should add a warning.

handibles:
Gentle post-yule pre-student bump.

antagomir:
[...]

PR #619, opened 2024-02-12
Matrix completion for robust CLR in decostand

antagomir:
This is intended to solve #570 

I am opening the PR in order to request feedback before final checks & possible polishing.

Do the maintainers see any shortcomings / issues to be addressed on this?

The matrix completion step is straighforward. The code has been modified from the ROptSpace package. The code was moved to vegan in order to avoid new dependencies, although using the actual original package directly would simplify this and I will happy to change into that if this idea receives support. The ROptSpace package has been cited and its license (MIT) is compatible with vegan's license (GPL) and the code from ROptSpace can be included in vegan. 

Unit tests and examples have been updated, too.

Comparisons with an independent Python implementation (deicode) are ongoing, I will check those as well to make sure that this works as expected.

Ah, should the PRs be done againsta vegandevs::master, or another branch?


DESCRIPTION
jarioksa, line 1: Please no version upgrades except with releases (that is looming after this is merged, vegan3d released + vegan upgraded for this, and some other pending fixes). So still 2.6-5, and the CRAN release would be 2.6-6.

R/decostand.R
```diff
@@ -260,3 +277,348 @@
         x[abs(x) < sqrt(.Machine$double.eps)] <- 0
     x
 }
+
+
+
+
+
+# .OptSpace : an algorithm for matrix reconstruction from a partially revealed set
```
jarioksa, line 285: This is not necessary, but I'd prefer doubled ## for comments on their own line.

R/decostand.R
```diff
@@ -260,3 +277,348 @@
...
+# \item{S}{an \eqn{(r \times r)} matrix as singular values.}
+# \item{Y}{an \eqn{(m \times r)} matrix as right singular vectors.}
+# \item{dist}{a vector containing reconstruction errors at each successive iteration.}
+# }
+# @references
+# Raghunandan H. Keshavan, Andrea Montanari, Sewoong Oh (2010).
+# Matrix Completion From a Few Entries.
+# IEEE Transactions on Information Theory 56(6):2980--2998.
+#
+.OptSpace <- function(A, ropt=NA, niter=50, tol=1e-6, showprogress=FALSE){
```
jarioksa, line 315: Other vegan functions uses trace = FALSE instead of showprogress. Wouldn't hurt here, although not crucial. These functions also usually cat() tracing instead of message().

Vegan usually has opening { on a new line in function definition (again: not crucial).

R-ext manual says that stop() (and warning()) messages should be seen as a part of a sentence and should not have period / full stop (.) at the end.

R/decostand.R
```diff
@@ -260,3 +277,348 @@
...
+      stop("* .OptSpace: ropt should be an integer in [1,min(nrow(A),ncol(A))].")
+    }
+  }
+  ## Preprocessing : niter : maximum number of iterations
+  if ((is.infinite(niter))||(niter<=1)||(!is.numeric(niter))){
+    stop("* .OptSpace: invalid niter number.")
+  }
+  niter = round(niter)
+  
+  m0 = 10000
```
jarioksa, line 360: Again nitpicking: I have followed C convention where arbitrary constants are written uppercase (M0, RHO).

R/decostand.R
```diff
@@ -260,3 +277,348 @@
...
+  if (showprogress){
+    message("* .OptSpace: Step 4: Gradient Descent ...")
+  }
+  X = X0
+  Y = Y0
+  S = .aux_getoptS(X,Y,M_E,E)
+  
+  # initialize
+  dist = array(0,c(1,(niter+1)))
+  dist[1] = norm((M_E - (X%*%S%*%t(Y)))*E,'f')/sqrt(nnZ.E)
```
jarioksa, line 419: Formatting again: use spaces around %*% for readability (still quite OK here, but becomes awkward when you have list elements with $-signs.

tests/aitchison-tests.R
```diff
@@ -34,6 +35,10 @@ a1 <- vegan::vegdist(testdata, method = "aitchison", pseudocount=1)
 a2 <- vegan::vegdist(testdata+1, method = "aitchison")
 max(abs(a1-a2)) < 1e-6 # Tolerance
 
+# Robust rclr+Euclid should give same result than robust.aitchison
+d1 <- vegan::vegdist(vegan::decostand(testdata, "rclr"), "euclidean")
```
jarioksa, line 39: In vegan, do not use vegan::function(): Ripley can catch this and complain (that has happened to me).

jarioksa (commented):
This is a first go-through and only concerns style. I'll test the working of the function later. However, looks fine at the first look, and I'll head for a CRAN release soon after this is merged.

jarioksa:
I'm heading for a CRAN release. It will still take some time: tests are ahead. In good case could be before May Day. Would this suit your plans with CLR?

antagomir:
The main bottleneck currently is that we cannot compare with external Python implementation to make sure the results are comparable. We were waiting for that (external issue) to become fixed. It would be good to get this fixed, I can try to see this week if this could be doable nevertheless. We will report back this week how it looks. Otherwise, it can wait until the next CRAN release.

antagomir:
This PR will be updated & reconsidered after #667 is ready.

If we add the matrix completion step in vegan this would be require a new dependency, the "filling" package, which has additional dependencies (CVXR, Rcpp, Rdpack, ROptSpace, RSpectra, nabor, stats, utils). 

Then the question is whether such new dependency would be acceptable?

Entirely reimplementing the filling operation in vegan might not be feasible but we can have a look if it is not possible add new dependencies.

antagomir:
Thanks for the feedback. We are making some final checks and I'm hoping to update & finalize this shortly now.

R/decostand.R
antagomir: done

R/decostand.R
antagomir: message changed to cat

openings updated

periods removed

R/decostand.R
antagomir: added

R/decostand.R
antagomir: true, added

antagomir:
I am closig this PR. I have taken into account the feedback on this PR and merged the relevant material in PR #667 in the hope that we can finish everything on one go soon.

PR #648, opened 2024-05-15
Add website to DESCRIPTION + move and tweak news to be pkgdown friendly.

olivroy:
The goal of having the website is enhanced discoverability and linking  https://blog.r-hub.io/2019/12/10/urls/

Moving NEWS.md to top level and changing version heading to first level should get the changelog automatically on the website.

https://pkgdown.r-lib.org/reference/build_news.html


gavinsimpson:
One thing I am slightly worried about is having documentation for the in-development version of *vegan*. That's partly why I haven't gone further down this route after getting the basic build up an running.

jarioksa:
Personally I'm absolutely non-interested on pkgdown. It was suggested long time ago, and some of my *young* co-developers implemented pkgdown. When I look at the pkgdown page, I see no value added (and  – fortunately – no VAT paid even if it is a private for-profit organization https://posit.co). In contrast, R-Universe provides a real service providing  binary installs of vegan and a better front page than pkgdown. Would *I* add a new URL to DESCRIPTION (which already has an URL), I'd prefer R-Universe that has the same information as the commercial pkgdown *plus* adds advantage of providing binary builds. However, I have no keen interest on the issue and I trust Gavin Simpson. Some minor issues follow: R manual for packages says that that *preferred* location of NEWS is inst/NEWS.  The manual allows NEWS in the main directory, but inst/NEWS takes preference. R manual also says that only Rd format is handled. I was aware of this when I switched from preferred format to .md just because utils::news still works except that it cannot show changes in *single* version. However, I thought that the thing that people *want* to see and *need* to see are the latest changes – who wants to see what happened in vegan 2.4-3? (It was possible while I had  Rd formatted NEWS). What your PR suggests is to give back that choice that I had before switching from Rd to md format. Then I thought that this loss was meaningless and now I think that the same gain would be useless. Anyway, I'm just non-interested and I leave the issue for the judgment of younger developers.

jarioksa:
I have no problems with this PR. I am not keenly interested in it either, but I just don't care. I didn't have any particular reason for using the header levels except that in some renditions toplevel (#) and second level (##) looked disproportionally big. That of course depends on the rendition. It seems that CRAN does more nicely with # and ## levels than my old choice. Neither do I object for moving NEWS to the "toplevel". My original choice was based of inst/ was based on the R extension manual, but that allows "toplevel" (even though  "toplevel" will be written over if there are inst/NEWS). Neither do I have much against adding that a new  URL as a homepage even if I really do think that this page is pretty useless and annoyingly juvenile (mickeysofty). Tthere is a lot of space in DESCRIPTION to use. However, I have updated inst/NEWS.md (and DESCRIPTION as ell) so much that currently this PR cannot be applied directly. Two alternatives: either you update your PR or I just use good old sed (ask your father) to change the heading levels and we can see other changes separately.

olivroy:
I have fixed the merge conflicts

PR #667, opened 2024-06-30
Fix to na.rm in decostand methods alr/clr 

### LINKED ISSUE #661: CLR: na.rm is not working

antagomir:
This fixes #661 

The `na.rm` option has been enabled in decostand methods "alr" and "clr".

Some notes:

1) "clr": With `na.rm=TRUE`, the NA entries in original data are replaced by 0 in the clr-transformed data; here zero is an arbitrary choice but no better solutions are available afaik, and the same approach is used in `compositions::clr` so at least we are in sync. With `na.rm=FALSE`, the entries that are NA in the original data will be also NAs in the clr-transformed data; the NAs are otherwise ignored in the calculations during the transformation. In the alternative implementation, `compositions::alr` NAs are always replaced with 0 (no na.rm option available). As far as I can see it is feasible to provide both options for `na.rm` for the "clr" method, we are providing that now.
 
2) "alr": similarly done for decostand "alr" transformation, user can choose whether to keep the NAs (`na.rm=FALSE`) or replace them by 0 (`na.rm=TRUE`). In comparison, I noticed that in the alternative implementation, `compositions::alr` NAs are always kept as NAs (no na.rm option available). As far as I can see it is feasible to provide both options for `na.rm` for the "alr" method, we are providing that now.

3) The "rclr" method will be handled separately in #619 

@TuomasBorman can you test and confirm that these are as you expected?


R/decostand.R
```diff
@@ -156,61 +156,93 @@
...
-    means <- rowMeans(clog, na.rm = na.rm)
+
+    # Calculate sample-wise log means (note: here we always set na.rm=TRUE !)
+    means <- rowMeans(clog, na.rm = TRUE)
+
+    # CLR transformation
     clog <- clog - means
+
+    # Replace missing values with 0
+    if (na.rm) {
```
TuomasBorman, line 183: You could add

`any(is.na(clog))`

as this is not affecting the result otherwise and the message could confuse

R/decostand.R
```diff
@@ -225,6 +257,13 @@
     clog <- log(x)
     refvector <- clog[, reference]
     clog <- clog[, -reference] - refvector
+
+    # Replace missing values with 0
+    if (na.rm) {
```
TuomasBorman, line 262: Same here, consider adding `any(is.na(clog))`

R/decostand.R
```diff
@@ -156,61 +156,93 @@
...
+    function (x, na.rm, pseudocount = 0, reference = 1)
 {
     # Add pseudocount
     x <- x + pseudocount
     # If there is negative values, gives an error.
-    if (any(x < 0, na.rm = na.rm)) {
-        stop("'alr' cannot be used with negative data: use pseudocount >= ",
+    # Always na.rm=TRUE at this step
+    if (any(x <= 0, na.rm = TRUE)) {
+        stop("'alr' cannot be used with non-positive data: use pseudocount >= ",
```
TuomasBorman, line 245: "use pseudocount >=". This is wrong. It should be "use pseudocount >"

TuomasBorman (approved):
I think now works as expected. See the comments.

R/decostand.R
```diff
@@ -156,61 +156,93 @@
 
 ## Modified from the original version in mia R package
 .calc_clr <-
-    function(x, pseudocount=0, na.rm = TRUE)
+    function(x, na.rm, pseudocount=0)
 {
```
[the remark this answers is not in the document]
antagomir: done

R/decostand.R
```diff
@@ -156,61 +156,93 @@
...
+   #} 
+
+   return(xx)
+   
 }
 
 
+
+
 .calc_alr <-
```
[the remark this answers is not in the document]
antagomir: ok

R/decostand.R
antagomir: Good catch.

antagomir:
Great, fixed the remarks.

antagomir:
This PR implements updated versions for:
- Robust CLR `decostand(..., method="rclr")`
- Robust Aitchison distance `vegdist(..., method="robust.aitchison")`

This PR is now updated with the following:
- Earlier separate PR #619 has been merged here, taking into account the feedback there (I hope the formatting issues are now ok, those seemed to be the main remarks)
- Approach is now synced with the Python implementation
- Handling of missing values has been improved.


One major point that I wanted to check before giving the very final touch on this PR: can we export the new function ROptSpace? This is not only used to calculate the robust.aitchison distance in vegdist but it is also more widely applicable and this would make the methods directly comparable with their equivalents in Python/gemelli, supporting cross-testing etc.

The OptSpace function is used for matrix completion after the rclr transformation. The rclr transformation usually results in some NA values in the transformed matrix, and the matrix completion is commonly done after this step. The reason for separating the rclr transformation and the matrix completion for missing values is that the matrix completion is not used on the rclr-transformed matrix directly; instead it returns other (SVD-related) outputs that are needed in further downstream analyses (e.g. robust Aitchison distance, robust PCA). The OptSpace matrix completion is used in the robust.aitchison distance in vegdist. 

The new OptSpace function is a modified version of ROptSpace::OptSpace function, with modifications to better fit ecological data. The original implementation has MIT license and it has been cited properly. This is compatible with the GPL2 license of vegan and can be included this way. 

So I would like to confirm if it is OK to keep the OptSpace function here I can make the final checks, otherwise I would remove it before declaring the PR complete.

antagomir:
Some minor updates added. 

This PR accommodates all previous feedback. 

It is ready and preliminary tests against the Python/Gemelli implementation have been done and the results are aligned.

If possible, we hope to keep `OptSpace` as an exported function as it has further use (e.g. robust PCA).

cameronmartino:
Hi,

I am the author of the method being added here & the python implementation (diecode now moved to [gemelli](https://github.com/biocore/gemelli)). Greatly appreciate @antagomir adding this implementation in R! I have been in some conversation with him on this offline. To help push the PR along, I checked the implementation in this PR and the results are now the same as the python implementation (as pointed out in issue #570 they did not match before). Simple test example below. Hope this helps get this PR through, there has been the demand in R but I didn't have the bandwidth to tackle this myself.

First, grab some (small) real data as an example
```bash
wget https://github.com/biocore/gemelli/raw/refs/heads/master/ipynb/tutorials/qiime2-moving-pictures-tutorial/table.biom
# export to tsv for R
biom convert -i table.biom -o table.tsv --to-tsv
sed -i '' 1d table.tsv 
````

Generate the distances and ordination in R (using this PR branch)

```R
library(devtools)
install_github("antagomir/vegan", ref = "nafix")
library(vegan)
# load table
table <- read.table("table.tsv", header = T, comment = '', check = F, sep = '\t', row.names = 1)
table <- t(as.matrix(table))
rclr_table <- vegan::decostand(table, method="rclr")
# calculate the distance
distance = vegan::vegdist(rclr_table, method = "euclidean")
write.table(as.matrix(distance), file = "R-rpca-distances.tsv", row.names=rownames(table), col.names=rownames(table), sep="\t")
# save to test RPCA
svd.res <- svd(rclr_table)
write.table(svd.res$u, file = "R-rpca-sample-loadings.tsv", row.names=rownames(table), sep="\t")
write.table(svd.res$d, file = "R-rpca-eig-vals.tsv", row.names=FALSE, sep="\t")
write.table(svd.res$v, file = "R-rpca-feature-loadings.tsv", row.names=colnames(table), sep="\t")
```

Generate the distances and ordination in python.

```python
import numpy as np
import pandas as pd
from biom import load_table
from gemelli.rpca import rpca
from gemelli import __version__
print('Version: %s' % __version__)

# load table
bt = load_table('table.biom')
# run RPCA (this runs RCLR internally)
py_ordination, py_distance = rpca(bt)
# extract RPCA results from skbio type
py_rpca_U = py_ordination.samples.values
py_rpca_V = py_ordination.features.values
py_rpca_S = py_ordination.eigvals.values
py_dist = py_distance.to_data_frame().values
# check the OptSpace results
R_rpca_U = pd.read_csv('R-rpca-sample-loadings.tsv', sep='\t', index_col=None).values[:, :3]
R_rpca_S = pd.read_csv('R-rpca-eig-vals.tsv', sep='\t', index_col=None).values[:3]
R_rpca_V = pd.read_csv('R-rpca-feature-loadings.tsv', sep='\t', index_col=None).values[:, :3]
R_rpca_dist = pd.read_csv('R-rpca-distances.tsv', sep='\t', index_col=None).values
# check
print(' Sample loadings match: %s \n' % str(abs((np.abs(R_rpca_U) - np.abs(py_rpca_U)).sum()) < 1e-6),
        'Feature loadings match: %s \n' % str(abs((np.abs(R_rpca_V) - np.abs(py_rpca_V)).sum()) < 1e-6),
        'Singular values match: %s \n' % str(abs((np.abs(R_rpca_S) - np.abs(py_rpca_S)).sum()) < 1e-6),
        'Distances match: %s \n' % str(abs((R_rpca_dist - R_rpca_dist).sum()) < 1e-6))
```
```output
Version: 0.0.12
 Sample loadings match: True 
 Feature loadings match: True 
 Singular values match: True 
 Distances match: True 
```

antagomir:
This is now thoroughly tested and cross-validated with an independent implementation and as far as I can see, we have fixed all related issues. Would the start of 2025 be a good time to merge the PR? ping @jarioksa

antagomir:
Dear @jarioksa @gavinsimpson - there has been general interest towards the robust CLR and robust Aitchison. These methods are already in vegan, following our previous PR. 

This present PR includes fixes to the previously reported issue + improved implementation, and it has been extensively verified with an independent Python implementation by @cameronmartino

If there are any issues remaining towards accepting this improved update in vegan we will be happy to work through them. As far as I can see, all previous feedback has been incorporated.

R/decostand.R
```diff
@@ -269,3 +311,323 @@
...
+  n <- nrow(x)
+  m <- ncol(x)
+  
+  ## Preprocessing : other sparse-related concepts
+  nnZ.E <- sum(!idxna)
+  E <- array(0, c(nrow(x), ncol(x))); E[!idxna] <- 1
+  eps <- nnZ.E/sqrt(m*n)
+  
+  ## Preprocessing : ROPT  : implied rank
+  if (is.na(ROPT)){
```
gavinsimpson, line 346: I don't like this design; Setting an argument to be `NA` implies that its value is *missing*. Better conventions are `NULL`. However, in this case, I think we should define `ropt = TRUE` as the argument and default value. Then in the function we would do

```r
if (ropt) {
  rpot <- 3 # set your default
}
```

and then we can document `ropt` to take a logical `TRUE/FALSE` or a positive integer for the rank.

R/decostand.R
```diff
@@ -269,3 +311,323 @@
...
+    if (verbose){
+      cat("* OptSpace: Guessing an implicit rank.")
```
gavinsimpson, line 348: This should be `message()`

R/decostand.R
```diff
@@ -269,3 +311,323 @@
...
+    }
+    r <- min(max(round(.guess_rank(M_E, nnZ.E)), 2), m-1)
+    if (verbose){
+      cat(paste0('* OptSpace: Guessing an implicit rank: Estimated rank : ',r))
```
gavinsimpson, line 352: `message()` not `cat()`

R/decostand.R
```diff
@@ -269,3 +311,323 @@
...
+    }
+  } else {
+    r <- round(ROPT)
+    if ((!is.numeric(r)) || (r<1) || (r>m) || (r>n)){
```
gavinsimpson, line 356: spacing around operators and before `{`

R/decostand.R
```diff
@@ -269,3 +311,323 @@
...
+      stop("* OptSpace: ROPT should be an integer in [1,min(nrow(x),ncol(x))]")
+    }
+  }
+  
+  ## Preprocessing : NITER : maximum number of iterations
+  if ((is.infinite(NITER))||(NITER<=1)||(!is.numeric(NITER))){
```
gavinsimpson, line 362: spacing around everything

R/decostand.R
```diff
@@ -269,3 +311,323 @@
...
+    stop("* OptSpace: invalid NITER number")
+  }
+  NITER <- round(NITER)
+  
+  m0 <- 10000
+  rho <-  eps*n
+  
+  ## Main Computation
+  rescal_param <- sqrt(nnZ.E*r/(norm(M_E,'f')^2))
```
gavinsimpson, line 371: Spacing

R/decostand.R
```diff
@@ -269,3 +311,323 @@
...
+  dist <- array(0, c(1, (NITER+1)))
+  dist[1] <- norm((M_E - (X %*% S %*% t(Y)))*E, 'f') / sqrt(nnZ.E)
+  for (i in seq_len(NITER)){
+    # compute the gradient
+    tmpgrad <- .aux_gradF_t(X, Y, S, M_E, E, m0, rho)
+    W <- tmpgrad$W
+    Z <- tmpgrad$Z
+    # line search for the optimum jump length
+    t <- .aux_getoptT(X, W, Y, Z, S, M_E, E, m0, rho)
+    X <- X + t*W;
```
gavinsimpson, line 410: no trailing `;`

R/decostand.R
```diff
@@ -269,3 +311,323 @@
...
+    Y <- Y + t*Z;
```
gavinsimpson, line 411: no trailing `;`

R/decostand.R
```diff
@@ -269,3 +311,323 @@
...
+  index_order <- order(diag(S), decreasing = TRUE)
+  X <- X[, index_order]
+  Y <- Y[, index_order]
+  S <- S[index_order, index_order]
+  out$X <- X
+  out$S <- S
+  out$Y <- Y
+  out$dist <- dist
+  if (verbose){
+    cat('* OptSpace: estimation finished.')
```
gavinsimpson, line 434: `message()`

R/decostand.R
```diff
@@ -269,3 +311,323 @@
...
+    }
+  }
+  
+  ## Preprocessing : NITER : maximum number of iterations
+  if ((is.infinite(NITER))||(NITER<=1)||(!is.numeric(NITER))){
+    stop("* OptSpace: invalid NITER number")
+  }
+  NITER <- round(NITER)
+  
+  m0 <- 10000
```
gavinsimpson, line 367: what is this constant for?

R/decostand.R
```diff
@@ -269,3 +311,323 @@
...
+# Aux 2 : compute the distortion ------------------------------------------
+# @keywords internal
+.aux_G <- function(X, m0, r)
+{
+  z <- rowSums(X^2)/(2*m0*r)
+  y <- exp((z-1)^2) - 1
+  idxfind <- (z < 1)
+  y[idxfind] <- 0
+  out <- sum(y)
+  return(out)
```
gavinsimpson, line 536: It is better style now to not use `return` as R automagically calls `return()` on the value of the last line of a function. Just leave this as `out`.

R/decostand.R
```diff
@@ -269,3 +311,323 @@
...
+# @keywords internal
+.aux_F_t <- function(X, Y, S, M_E, E, m0, rho)
+{
+  n <- nrow(X)
+  r <- ncol(X)
+  out1 <- (sum((((X %*% S %*% t(Y)) - M_E)*E)^2))/2
+  out2 <- rho*.aux_G(Y,m0,r)
+  out3 <- rho*.aux_G(X,m0,r)
+  out  <- out1+out2+out3
+  return(out)
```
gavinsimpson, line 548: It is better style now to not use `return` as R automagically calls `return()` on the value of the last line of a function. Just leave this as `out`.

R/decostand.R
```diff
@@ -269,3 +311,323 @@
...
+  M <- as.matrix(scale(M, center=TRUE, scale=FALSE))
+  # Center rows to 0
+  M <- as.matrix(t(scale(t(M), center=TRUE, scale=FALSE)))
+
+  # Imputed matrix; add to the output
+  out$M <- M
+
+  # -------------------------------------------
+
+  return(out)
```
gavinsimpson, line 457: It is better style now to not use `return` as R automagically calls `return()` on the value of the last line of a function. Just leave this as `out`.

gavinsimpson, line 457: It is better style now to not use `return` as R automagically calls `return()` on the value of the last line of a function. Just leave this as `out`.

R/decostand.R
```diff
@@ -156,61 +156,96 @@
...
+     M_I <- as.matrix(scale(M_I, center=TRUE, scale=FALSE))
+
+     # Center rows to 0
+     M_I <- as.matrix(t(scale(t(M_I), center=TRUE, scale=FALSE)))
+
+     # Imputed matrix
+     xx <- M_I
+   }
+  
+   return(xx)
```
gavinsimpson, line 236: It is better style now to not use `return` as R automagically calls `return()` on the value of the last line of a function. Just leave this as `out`.

R/decostand.R
```diff
@@ -269,3 +311,323 @@
...
+  Qx <- ((t(X) %*% ((M_E-XSY)*E) %*% YS)/n)
+  Qy <- ((t(Y) %*% t((M_E-XSY)*E) %*% XS)/m)
+  
+  W <- (((XSY-M_E)*E) %*% YS)  + (X %*% Qx) + rho*.aux_Gp(X, m0, r)
+  Z <- (t((XSY-M_E)*E) %*% XS) + (Y %*% Qy) + rho*.aux_Gp(Y, m0, r)
+  
+  resgrad <- list()
+  resgrad$W <- W
+  resgrad$Z <- Z
+  return(resgrad)
```
gavinsimpson, line 585: Just `resgrad`.

R/decostand.R
```diff
@@ -269,3 +311,323 @@
...
+    for (j in seq_len(r)){
+      ind <- ((j-1)*r+i)
+      tmp <- t(X) %*% (outer(X[,i], Y[,j])*E) %*% Y      
+      A[,ind] <- as.vector(tmp)
+    }
+  }
+  
+  S <- solve(A, C)
+  out <- matrix(S, nrow=r)
+  return(out)
```
gavinsimpson, line 611: `out`

R/decostand.R
```diff
@@ -269,3 +311,323 @@
...
+}
+
+# Aux 5 : optimal line search ---------------------------------------------
+# @keywords internal
+.aux_getoptT <- function(X, W, Y, Z, S, M_E, E, m0, rho)
```
gavinsimpson, line 616: These arguments need comments to help us understand what they are for - will help with ongoing maintenance of the code

R/decostand.R
```diff
@@ -269,3 +311,323 @@
...
+  resgrad$W <- W
+  resgrad$Z <- Z
+  return(resgrad)
+  
+}
+
+
+# Aux 4 : Sopt given X and Y ----------------------------------------------
+# @keywords internal
+.aux_getoptS <- function(X, Y, M_E, E)
```
gavinsimpson, line 592: These arguments need comments to help us understand what they are for - will help with ongoing maintenance of the code

R/decostand.R
```diff
@@ -269,3 +311,323 @@
...
+.aux_Gp <- function(X,m0,r)
+{
+  z <- rowSums(X^2)/(2*m0*r)
+  z <- 2*exp((z-1)^2)/(z-1)
+  idxfind <- (z<0)
+  z[idxfind] <- 0  
+  out <- (X * matrix(z, nrow=nrow(X), ncol=ncol(X), byrow=FALSE))/(m0*r)
+}
+# @keywords internal
+.aux_gradF_t <- function(X, Y, S, M_E, E, m0, rho)
```
gavinsimpson, line 563: These arguments need comments to help us understand what they are for - will help with ongoing maintenance of the code

R/decostand.R
```diff
@@ -269,3 +311,323 @@
...
+  out2 <- rho*.aux_G(Y,m0,r)
+  out3 <- rho*.aux_G(X,m0,r)
+  out  <- out1+out2+out3
+  return(out)
+}
+
+
+# Aux 3 : compute the gradient --------------------------------------------
+# @keywords internal
+.aux_Gp <- function(X,m0,r)
```
gavinsimpson, line 554: These arguments need comments to help us understand what they are for - will help with ongoing maintenance of the code.

And spaces after `,`

R/decostand.R
```diff
@@ -269,3 +311,323 @@
...
+  for (i in seq_len(21)){
+    f[i+1] <- .aux_F_t(X+t*W, Y+t*Z, S, M_E, E, m0, rho)
+    if ((f[i+1]-f[1]) <= 0.5*t*norm2WZ){
+      out <- t
+      break
+    }
+    t <- t/2
+  }
+  out <- t
+  return(t)
```
gavinsimpson, line 631: `t`

R/vegdist.R
```diff
@@ -48,7 +48,7 @@
     if (method == 21)  # aitchison
         x <- decostand(x, "clr", ...)  # dots to pass possible pseudocount
     if (method == 22)  # robust.aitchison
-        x <- decostand(x, "rclr") # No pseudocount for rclr
+        x <- decostand(x, "rclr", nr.rm=na.rm, ...) # No pseudocount for rclr
```
gavinsimpson, line 51: Is this a typo? Should it be `na.rm = na.rm, ...`?

man/OptSpace.Rd
```diff
@@ -0,0 +1,77 @@
+\encoding{UTF-8}
+\name{OptSpace}
+\alias{OptSpace}
+\title{OptSpace: algorithm for matrix reconstruction from a partially revealed set}
+\description{
+This function was adapted from the original source code in the
+ROptSpace R package (version 0.2.3; MIT License) by Raghunandan
```
gavinsimpson, line 7: Use `\pkg{package_name}` to markup an R package name

man/OptSpace.Rd
```diff
@@ -0,0 +1,77 @@
...
+H. Keshavan, Andrea Montanari, Sewoong Oh (2010). See
+ROptSpace::OptSpace for more information.  Let's assume an ideal
```
gavinsimpson, line 9: `ROptSpace::OptSpace` -> `\code{ROptSpace::OptSpace}`, assuming we don't want to link to the function in the package (if we do then the package would have to be `Suggests`, which I think we wanted to avoid?)

man/OptSpace.Rd
```diff
@@ -0,0 +1,77 @@
...
+matrix \eqn{M} with \eqn{(m\times n)} entries with rank \eqn{r} and we
+are given a partially observed matrix \eqn{M\_E} which contains many
+missing entries. Matrix reconstruction - or completion - is the task
+of filling in such entries. OptSpace is an efficient algorithm that
+reconstructs \eqn{M} from \eqn{|E|=O(rn)} observed elements with
+relative root mean square error (RMSE) \deqn{RMSE \le
+C(\alpha)\sqrt{nr/|E|}}.
+}
+\usage{
+OptSpace(x, ROPT=3, NITER=5, TOL=1e-5, verbose=FALSE)
```
gavinsimpson, line 19: update these args when the function switches to lower case argument names

man/OptSpace.Rd
```diff
@@ -0,0 +1,77 @@
...
+  Knight, R. & Zengler, K. (2019) A novel sparse compositional technique
+  reveals microbial perturbations.
+  \emph{mSystems} \strong{4}, 1.
+
+  }
+
+\examples{
+data(varespec)
+# rclr transformation with no matrix completion for the 0/NA entries
+x <- decostand(varespec, method="rclr", impute=FALSE)
```
gavinsimpson, line 72: spaces around `=`

man/OptSpace.Rd
```diff
@@ -0,0 +1,77 @@
...
+# Add matrix completion
+xc <- OptSpace(x, ROPT=3, NITER=5, TOL=1e-5, verbose=FALSE)$M
```
gavinsimpson, line 74: spaces around `=`

man/decostand.Rd
```diff
@@ -106,25 +106,37 @@ decobackstand(x, zap = TRUE)
      (e.g. the smallest positive value in the data), either by
      adding it manually to the input data, or by using the argument
      \code{pseudocount} as in
-     \code{decostand(x, method = "clr", pseudocount = 1)}. Adding
+     \code{decostand(x, method = "clr", na.rm=TRUE, pseudocount = 1)}. Adding
```
gavinsimpson, line 109: `na.rm = TRUE,`

man/decostand.Rd
```diff
@@ -106,25 +106,37 @@ decobackstand(x, zap = TRUE)
...
      \deqn{rclr = log\frac{x}{g(x > 0)}}{%
-     rclr = log(x/g(x > 0))}
-     where \eqn{x} is a single value, and \eqn{g(x > 0)} is the geometric 
-     mean of sample-wide values \eqn{x} that are positive (> 0).
+           rclr = log(x/g(x > 0))} where \eqn{x} is
+     a single value, and \eqn{g(x > 0)} is the geometric mean of
+     sample-wide values \eqn{x} that are positive (> 0). The OptSpace
+     algorithm performs matrix completion for the missing values
+     that result from log transformation of the zero entries in the
+     original input data. The vegan implementation of the matrix
```
gavinsimpson, line 127: I don't think we need to repeat this
>The vegan implementation of the matrix
     completion was adapted from the original implementation
     ROptSpace::OptSpace (version 0.2.3) following Keshavan et
     al. (2010).

here. Just direct people to the OptSpace Rd file with `See \link[vegan]{OptSpace} for more details.` Or something like that.

man/decostand.Rd
```diff
@@ -204,13 +220,17 @@ sptrans <- wisconsin(varespec)
 # CLR transformation for rows, with pseudocount
 varespec.clr <- decostand(varespec, "clr", pseudocount=1)
 
+# Robust CLR (rclr) transformation for rows, no pseudocount necessary
+varespec.rclr <- decostand(varespec, "rclr", impute=TRUE)
```
gavinsimpson, line 224: `impute = TRUE`

man/vegdist.Rd
```diff
@@ -419,5 +419,8 @@ vare.dist <- vegdist(decostand(varespec, "norm"), "euclidean")
 vare.dist <- vegdist(decostand(varespec, "log"), "altGower")
 # Range standardization with "altGower" (that excludes double-zeros)
 vare.dist <- vegdist(decostand(varespec, "range"), "altGower")
+# Robust Aitchison distance equals to Euclidean distance for rclr transformed data
+vare.dist <- vegdist(decostand(varespec, "rclr"), method="euclidean")
```
gavinsimpson, line 423: `method = "euclidean"`

tests/decostand-tests.R
gavinsimpson, line 1: Ideally, this file would have similar standards to the rest of vegan, with spacing around operators, like `=`, etc. But I don't think it needs doing personally. @jarioksa may differ.

gavinsimpson (changes requested):

gavinsimpson:
@antagomir I have done a code review - mostly trivial for style, but matching vegan-style is needed to help us maintain code. I found one typo-bug. Some comments and suggestions to change the code and documentation in places.

R/decostand.R
```diff
@@ -156,61 +156,96 @@
...
+
+    # Calculate sample-wise log means (note: here we always set na.rm=TRUE !)
+    means <- rowMeans(clog, na.rm = TRUE)
+
+    # CLR transformation
     clog <- clog - means
+
+    # Replace missing values with 0
+    if (na.rm && any(is.na(clog))) {
+        cat("Replacing missing values with zero for clr. You can disable this with na.rm=FALSE.")
```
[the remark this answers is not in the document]
antagomir: Right - I was somehow mistaken that "cat" would be preferred in vegan. I agree, we go with `message()`.

Unless I am mistaken, the decostand default is na.rm=FALSE (in [decostand.R, line 2](https://github.com/vegandevs/vegan/blob/186573d038a6f7ece120a3d5af5e5eb2545146a6/R/decostand.R#L2))?

Agreed with warning and formatting & updated now.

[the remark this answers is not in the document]
gavinsimpson: > Unless I am mistaken, the decostand default is na.rm=FALSE

Right, sorry. Confused myself just scrolling through the diff.

tests/decostand-tests.R
antagomir: done

R/decostand.R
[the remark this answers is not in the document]
antagomir: Ok good. Warning is warranted in any case. Added.

antagomir:
Ok I have resolved all comments. I tried to check all files comprehensively for similar issues, I hope I covered all. 

The tests seem to pass ok.

Cameron Martino has substantially contributed to the implementation. I am not sure what the policy is but you can consider adding him in the DESCRIPTION file (or I can add it in PR if agreed so). 

Ready to merge? If anything else remains just let me know and we will do asap.

gavinsimpson:
@antagomir I don't see that my comments have been resolved - there are still trailing `;` in places, and using `return(foo)` as the last statement in a function is redundant, just use `foo`. Those were just a few things that I saw in `decostand.R` that are outstanding; I didn't check all the comments, those were ones that I noticed and decided to wait until they are complete before signing off.

R/decostand.R
```diff
@@ -269,3 +311,323 @@
...
+    x <- as.matrix(x)
+  }
+  if (!is.matrix(x)){
+    stop("* OptSpace : an input x should be a matrix")
+  }
+  if (any(is.infinite(x))){
+    stop("* OptSpace : no infinite value in x is allowed")
+  }
+  if (!any(is.na(x))){
+    stop("* OptSpace : there is no unobserved values as NA")
```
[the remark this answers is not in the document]
antagomir: true, good catch.

R/decostand.R
antagomir: Scaling parameter for gradient calculation. I added a small comment for this internal parameter. References for the method have been provided on the manpage.

R/decostand.R
antagomir: i added a short summary on all

R/decostand.R
antagomir: i added a short summary on all

R/decostand.R
antagomir: i added a short summary on all

R/vegdist.R
antagomir: fixed

antagomir:
Thanks!

I had resolved the comments as I went through and solved them. I have now reopened them for checking.

Github had hidden your 20+ comments that needed to be separately clicked open ("show more"). I had somehow not noticed that and mistakingly thought that I already went through all issues. Now I saw those, including trailing semicolon and foo etc. I have addressed all feedback now but there is still something that escaped my attention just let me know.

This PR is ready from my side but ready to add fixes as necessary.

antagomir:
The GHA fails but the issue seems unrelated to this PR - I also get the same "scores" error when I take a fresh clone from vegandevs/vegan master.

jarioksa:
@antagomir the error on missing `scores` **is** related to this PR: you add function `optspace`, but in `NAMESPACE` you export function `OptSpace`. Exporting a non-existent function ruins parsing the `NAMESPACE` exports and this first emerges when you try to build dynamic issues in help page `man/plot.cca.Rd`. Fixing the `NAMESPACE` will solve this issue.

antagomir:
Ahh, yes. The NAMESPACE of course.

Should be ok now?

gavinsimpson (approved):
There's some trivial things that I might clean up, but I can do that during the merge. Thanks @antagomir

tests/decostand-tests.R
```diff
@@ -1,50 +1,89 @@
...
-alt.clr <- t(apply(as.matrix(relative.with.pseudo), 1, FUN=function(x){
-log(x) - mean(log(x))}))
-max(abs(x.clr-alt.clr)) < 1e-6      
-all((x.rclr==0) == (testdata==0))
+alt.clr <- t(apply(as.matrix(relative.with.pseudo), 1, FUN = function(x){log(x) - mean(log(x))}))
+max(abs(x.clr - alt.clr)) < 1e-6      
+
+# Test that NAs are handled as expected in CLR
+x <- testdata
+x[sample(prod(dim(x)), 50)] <- NA # insert some NAs in the data
```
gavinsimpson, line 34: Should this have a seed set before we call `sample()`?

tests/decostand-tests.R
```diff
@@ -1,50 +1,89 @@
...
+# Works correctly also with other MARGIN
+inds <- !is.na(x) # Non-NA values
+max(abs(decostand(x, "clr", MARGIN = 2, na.rm = FALSE, pseudocount = 1)[inds] -
+        decostand(x, "clr", MARGIN = 2, na.rm = TRUE, pseudocount = 1)[inds])) < 1e-6
+# Results match for the non-NA values always (with tolerance 1e-6)
+inds <- !is.na(x) # Non-NA values
+max(abs(na.omit(decostand(x, "alr", na.rm = FALSE, pseudocount = 1)[inds] -
+                decostand(x, "alr", na.rm = TRUE, pseudocount = 1)[inds]))) < 1e-6
+
+# Test that NAs are handled as expected in ALR
```
gavinsimpson, line 56: I'm seeing problems with this section

tests/decostand-tests.R
```diff
@@ -1,50 +1,89 @@
...
+max(abs(x.clr - alt.clr)) < 1e-6      
+
+# Test that NAs are handled as expected in CLR
+x <- testdata
+x[sample(prod(dim(x)), 50)] <- NA # insert some NAs in the data
+# NAs in the original data remain NAs in the returned data
+all(is.na(decostand(x, "clr", na.rm = FALSE, pseudocount = 1)[is.na(x)])) == TRUE # NAs
+# For the other (non-NA) values, we get non-NA values back
+any(is.na(decostand(x, "clr", na.rm = FALSE, pseudocount = 1)[!is.na(x)])) == FALSE
+any(is.na(decostand(x, "clr", MARGIN = 2, na.rm = FALSE, pseudocount = 1)[!is.na(t(x))])) == FALSE
```
gavinsimpson, line 39: This is not returning `TRUE`:

```{r}
> any(is.na(decostand(x, "clr", MARGIN = 2, na.rm = FALSE, pseudocount = 1)[!is.na(t(x))])) == FALSE
[1] FALSE
```

Is this expected?

antagomir: Despite all the efforts to check each line, this had somehow escaped my attention.. good catch. This should be

`any(is.na(decostand(x, "clr", MARGIN = 2, na.rm = FALSE, pseudocount = 1)[!is.na(x)])) == FALSE` 

i.e. no transpose around `is.na(x)`.

antagomir: You can change, or I update the PR?

gavinsimpson: I'll change; I have a few more lines to check with you that are causing some output diffs for me but I won't get to those until later this evening or first thing tomorrow.

tests/decostand-tests.R
```diff
@@ -53,35 +92,52 @@ test2 <- test2[, -1]
 # Expect high correlation
 cor(unlist(test), unlist(test2)) > 0.99
```
gavinsimpson, line 93: The comment above suggests that we should see some `TRUE` values here but in my testing, all of these correlations are *not* $>0.99$. So I'm not sure what this section is supposed to be testing. It would be good to have an actual expectation to compare with, but clarifying whether something needs to be changed here would be a start?

tests/decostand-tests.R
```diff
@@ -53,35 +92,52 @@ test2 <- test2[, -1]
...
+
+# rclr transformation with no matrix completion for the 0/NA entries
+x2 <- decostand(varespec, method = "rclr", impute = FALSE)
+
+# Matrix completion
+x2c <- optspace(x2, ropt = 3, niter = 5, tol = 1e-5, verbose = FALSE)$M
+all(as.matrix(x1) == as.matrix(x2c))
+
+x2c <- optspace(x2, ropt = 1, niter = 5, tol = 1e-5, verbose = FALSE)$M
+all(as.matrix(x1) == as.matrix(x2c))
```
gavinsimpson, line 107: This is returning `FALSE` for me, but testing for exact equality is not very useful here. With `all.equal()`, I'm seeing

```r
> all.equal(as.matrix(x1), as.matrix(x2c))
[1] "Attributes: < Names: 1 string mismatch >"                                         
[2] "Attributes: < Component 2: Modes: list, numeric >"                                
[3] "Attributes: < Component 2: Length mismatch: comparison on first 2 components >"   
[4] "Attributes: < Component 2: Component 1: target is NULL, current is numeric >"     
[5] "Attributes: < Component 2: Component 2: Modes: character, numeric >"              
[6] "Attributes: < Component 2: Component 2: Lengths: 44, 1 >"                         
[7] "Attributes: < Component 2: Component 2: target is character, current is numeric >"
[8] "Mean relative difference: 0.9140407"
```

where the last line shows a significant deviation between `x1` and `x2c` - is this to be expected or not?

tests/decostand-tests.R
gavinsimpson: As you use `testdata` much later in L71 and L72 and dump the output to the test log we need to make this a fixed seed I think otherwise every run gives different results.

tests/decostand-tests.R
antagomir: This test was related to the earlier non-imputation version and can be removed.

gavinsimpson: Great - I'll make that change locally.

tests/decostand-tests.R
antagomir: This one (with ropt = 1); lines 106-107 can be removed. These rows might have been left there by accident as x1 and x2c are allowed differ here.

antagomir:
Ok I have responded now. My bad overlooking those two cases, I hope all ok now.

gavinsimpson:
Thanks @antagomir all merged now

jarioksa:
This was much bigger than I anticipated. Thanks for the hard work @antagomir and @gavinsimpson. Now only two things are missing: NEWS entry and updating authorship in DESCRIPTION.

It took a while to get this merged, but the main thing is how fast this goes to CRAN. I'm going to start working on a major CRAN release (2.7-x). There are a few things that must be solved, and I'm going to have a new issue on release plan.

Issue #677, opened 2024-08-13
Consider changing the default to `by = NULL` or `by = "margin"` in `adonis2()`


gavinsimpson:
The default permutation test in `adonis2()` is `by = "terms"` for a sequential test of model terms where order of terms can matter. I appreciate that this was the default (and only option in `adonis()`?) but it really isn't very useful in many situations that ecologists will be using `adonis2()` for (observational data).

`by = "margin"` would be better if you want a test of model terms by default. `by = NULL` would be more consistent with the `anova()` methods for `rda()`, `cca()`, etc.

I think we should change this to do the statistically better thing at the expense of backwards "compatability" with `adonis()`.

jarioksa:
Personally I think that `by = "terms"` is statistically more correct than `by = "margin"`: single terms sum up to total in sequential tests, but not in marginal tests. I only added that alternative to get rid of whining. That said, I agree with changing the default, but I'd prefer `by = NULL`.

gavinsimpson:
This is now implemented in the master branch and in the cran-2.6-6 branches.

jarioksa:
An unfortunate consequence of this change is that several packages depending on **vegan** manipulate and post-process `adonis2` result, and they assume that the result is from the old default `by = "terms"`. The fix is easy: just add explicit `by = "terms"`. However, we broke API, and there are many packages that need this fix. Should we reconsider this change and return to the old default behaviour?

Reverse dependency tests of 243 packages on April 9, 2025 found that these files used `adonis2`, but all should be inspected for the need of adding `by = "terms"`:
```sh
AICcPermanova/R/AICc_permanova2.R
AICcPermanova/R/fit_models.R
BiodiversityR/R/nested.npmanova.R
HDANOVA/R/wrappers.R
LorMe/R/structure_plot.R
MicrobiotaProcess/R/method-mp_adonis_anosim.R
MiscMetabar/R/beta_div_test.R
PCPS/R/pcps.sig.R
PCPS/R/select.pcpsmethod.R
ProtE/R/DIAnn.R
ProtE/R/max_quant.R
ProtE/R/new_vers_PD.R
ProtE/R/path_res.R
QFASA/R/All_Rep_Code.R
RVAideMemoire/R/pairwise.perm.manova.R
adespatial/R/stimodels.R
animalcules/R/diversity_beta_test.R
dispRity/R/adonis.dispRity.R
enmpa/R/niche_signal.R
mia/R/getPERMANOVA.R
[...]

PR #681, opened 2024-08-16
Change default test from sequential to omnibus in `adonis2()` Issue 677

gavinsimpson:
As per discussion in #677, this PR changes the default test from sequential (`by = "terms"`) to an omnibus test (`by = NULL`).

I updated the code, the Rd, the examples, and the test code/output.

I didn't update the example reference output here in this PR to keep it simple and there are other changes in master.

If accepted, consider when we include it in a CRAN release. This is a breaking change, and while I can justify this change in terms of making the code base more consistent across functions, it is *breaking*. Should this go into a 2.7.0 or wait for a 3.0.0?

Thoughts?

### DIFF
--- a/R/adonis.R
+++ b/R/adonis.R
@@ -1,6 +1,6 @@
 `adonis2` <-
     function(formula, data, permutations = 999, method = "bray",
-             sqrt.dist = FALSE, add = FALSE, by = "terms",
+             sqrt.dist = FALSE, add = FALSE, by = NULL,
              parallel = getOption("mc.cores"), na.action = na.fail,
              strata = NULL, ...)
 {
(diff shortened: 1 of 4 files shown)


jarioksa (approved):
This looks great. I think this should be merged to cran-2.6-6 as well (but I leave it to you to decide). This deserves a NEWS entry since it changes user experience, but that can be done separately.

gavinsimpson:
I've merged this and cherry-picked these 4 commits to the cran-2.6-6 branch and added a NEWS entry there. I should probably cherry pick that NEWS entry addition to the master branch too.

jarioksa:
cherry-picking won't work: cran-2.6-6 has `inst/NEWS.md` but master has `./NEWS.md`. I have only copied these entries. However, the current master NEWS must be edited for 2.6-6.2 release once that happens, and that will be very much manual work.

gavinsimpson:
Yeah; just found that out. I manually committed the news entry to master.

PR #729, opened 2025-02-24
make text.cca work when 'select' is a character vector

fpusan:
- `text.cca` was not plotting anything when `select` was a character vector, because it was not able to retrieve the labels for each point.
- I changed it to use a similar solution to that implemented in `text.metaMDS`, in which labels are taken directly from the point matrix returned by `.checkSelect`.

I'm not familiar with doing pull request to big repos, so kindly let me know if you need anything else.

### DIFF
--- a/R/text.cca.R
+++ b/R/text.cca.R
@@ -5,6 +5,15 @@
 {
     if (length(display) > 1)
         stop("only one 'display' item can be added in one command")
+
+    display <- match.arg(display, c("species", "sites", "wa", "lc", "bp", "cn", "reg", "all"))
+
+    if(missing(labels)) {
+        custom_labels <- FALSE
+    } else {
+        custom_labels <- TRUE
+    }
+
     pts <- scores(x, choices = choices, display = display, scaling = scaling,
                   const, correlation = correlation, hill = hill, tidy=FALSE,
                   droplist = TRUE)
@@ -12,12 +21,19 @@
     ## labels, the checks in "cn" branch fail and "bp" branch will
     ## be entered even if there should be no "bp" plotting
     cnam <- rownames(pts)
-    if (missing(labels))
+    if (!custom_labels)
         labels <- labels.cca(x, display)
     if (!missing(select)) {
         pts <- .checkSelect(select, pts)
-        labels <- rownames(pts)
+        if(custom_labels) {
+            if(class(select) %in% c("numeric", "logical") & length(labels) == length(cnam)) {
+                labels <- labels[select]
+            }
+        } else {
+            labels = rownames(pts)
+        }
     }
+    
     ## centroids ("cn") have special treatment: also plot biplot
     ## arrows ("bp") for continuous variables and ordered factors.
     if (display == "cn") {
--- a/R/text.cca.R
+++ b/R/text.cca.R
@@ -16,7 +16,7 @@
         labels <- labels.cca(x, display)
     if (!missing(select)) {
         pts <- .checkSelect(select, pts)
-        labels <- labels[select]
+        labels <- rownames(pts)
     }
     ## centroids ("cn") have special treatment: also plot biplot
     ## arrows ("bp") for continuous variables and ordered factors.


fpusan:
I can confirm it also works with "species", since it was my original use case.
From the way `pts` is generated, I just assumed it would contain the right data regardless of the choice of `display`
```
pts <- scores(x, choices = choices, display = display, scaling = scaling,
                  const, correlation = correlation, hill = hill, tidy=FALSE,
                  droplist = TRUE)
```
s

jarioksa:
@fpusan yes, I noticed this soon after writing my comment (that I deleted).

jarioksa:
It seems that `text.cca` is pretty complicated and has many options that do not work well together. One thing is that `labels`  can be used to replace the plotted text, but this does not work with your fix:
```r
data(dune, dune.taxon)
m <- cca(dune)
plot(m, type = "n")
text(m, display = "sites", labels = dune.taxon$Genus) # text is Genus name
text(m, display = "sites", labels = dune.taxon$Genus, select = c("Trifrepe","Trifprat") # text is 8-letter acronym
text(m, display = "sites", labels = dune.taxon$Genus, select = "Trifolium") # error (I don't know if this should be error or work
```
Without your fix, `select` won't work. However, setting `labels` should be allowed as well. I think this can be handled, but I don't have time to check it now.

There is also a problem in `labels.cca`:
```r
plot(m, type="n")
text(m, display = "sites") # OK
text(m, display = "site") # plots nothing: 'sites' cannot be abbreviated, fix obvious with match.arg()
```

I think we need a test set to see that all works like it should. I don't have time to think about it now. There are still cases with centroids, regression etc. that I haven't looked at.

jarioksa:
I'm on the move now, but all this sounds very good! I'll check this
tomorrow. Cheers, Jaz

ma 24. helmik. 2025 klo 16.55 Fernando Puente-Sánchez <
***@***.***> kirjoitti:

jarioksa:
You found a real problem in using `select`. However, `text.cca` is a kludgy function and it cannot be made any better by adding more kludges. It should be made simpler. Indeed, it can be made simpler because the following already works:
```r
data(dune)
mod <- cca(dune)
## your example: fails without your fix
plot(mod, type="n")
text(mod, display = "species", select = c("Trifrepe", "Trifprat"))
## works both with CRAN release 2.6-10 and  development version 2.7-0
plot(mod, type = "n") |> text("species", select = c("Trifrepe", "Trifprat"))
```
The crux is that with pipe (`|>`) we use `text.ordiplot`  instead of `text.cca`. So we should base `text.cca` on `text.ordiplot` and make it much more compact, streamlined and maintainable. This we have done with most other `text` (and `points`!) functions.

fpusan:
Thanks for looking into this!
I agree that refactoring `text.cca` is probably the best choice. However, I briefly looked at the source code for `text.ordiplot` and I am definitely not familiar enough with the internals of vegan to attempt this.
So I am glad I at least helped to identify the issue! :)

jarioksa:
@fpusan Thanks for bringing this to my attention. Honestly I had neglected these functions. The `plot.cca` function was completely re-designed with full flexibility and very versatile workflow was also made possible via pipes. I didn't see much use for these ole functions from 2003, and they were forgotten in the dark corners of the package. I even warned against there use in `plot.cca` help page! Now they are streamlined (I removed 65 lines of their code), and should work similarly as `text` and `points` functions that we use in pipes. However, this can be seen only after user experience: there may be things that I didn't see and that need inspection. Please try the functions and if any trouble, let me know!

fpusan:
Hi again!
Testing from 96f8f2e:

```
data(dune, dune.taxon)
m <- cca(dune)

plot(m, type = "n")
text(m, display = "species", select = c("Agrostol","Trifprat"), labels = c("foo", "bar") )
```

fails with

```

I think I am in the right commit, since the source code for `vegan:::text.cca` is much simplified.
Error in dimnames(x) <- dn : 
  length of 'dimnames' [1] not equal to array extent
```

jarioksa:
@fpusan Yes, this was intentional. Other modern functions require that `labels` match rows in the original data. **Vegan** is not quite consistent, but I'm just in the process of making it consistent. After I finish with this, I will open an issue where I discuss *changing* this behaviour for all functions similar to your suggestions. However, now we still are in the old mode. After possible comments on the future issue, I will decide which mode we'll adopt.

PR #742, opened 2025-04-22
Avoid memcpy of invalid pointer

MichaelChirico:
See discussion: https://github.com/r-lib/vctrs/pull/1968

Not found in any failing tests, so it's possible this code cannot be run because it'll never receive `nr | nc == 0` inputs.

It _might_ also be advisable to skip `memcpy()` where `src` is the `R_alloc()` output, I can't quite tell from the implementation whether it will return an invalid pointer when `N==0`.

### DIFF
--- a/src/nestedness.c
+++ b/src/nestedness.c
@@ -866,7 +866,7 @@ SEXP do_swap(SEXP x, SEXP nsim, SEXP thin, SEXP method)
 
     /* sequential trialswap of ix and save result to the iout
        array */
-    memcpy(ix, INTEGER(x), N * sizeof(int));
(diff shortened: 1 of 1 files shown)


jarioksa:
It really should never receive input `nr | nc == 0`, but you may want to be sure. However, I think it would be better to exit early with error if `N == 0` rather than proceed silently with a result that will trigger a cryptic error somewhere later. Just a quick first thought, and I'll have a look at this after a night's sleep.

jarioksa:
Yes, it is possible to call this code with zero-dim matrix, and this will segfault. I assume it is also possible to call (some of these) functions with a data which go to eternal loop. These need more extensive sanity checks before C calls.

Issue #785, opened 2026-04-20
ordistep scoping bug


PeetoomHeida:
I have had a lot of trouble using ordiR2step within a self-contained function. There seems to be a bug with the scoping of ordiR2step, where it will always look for the data object in the global environment, rather than respecting the local environment from which ordiR2step was called. I am running into this in the context of trying to parallelize some analyses, but it is easily run into in single-threaded contexts by embedding ordistep in any function. This applies to both ordistep and ordiR2step. See the MWE below:

```
# --- 1. This function fails ---
# This function replicates the 'object not found' error because 
# ordiR2step looks for 'env_data' in the parent/global environment
fail_rda_function <- function(comm, env) {
  # Local variables inside the function scope
  comm_mat <- comm
  env_data <- env
  
  mod0 <- rda(comm_mat ~ 1, data = env_data)
  mod1 <- rda(comm_mat ~ ., data = env_data)
  
  # This usually fails with "object 'env_data' not found"
  step_res <- ordiR2step(mod0, scope = formula(mod1), 
                         data = env_data, trace = FALSE)
  return(step_res)
}

# --- 2. THE WORKAROUND: Global Assignment ---
# This function succeeds by forcing the objects into the Global environment
success_rda_function <- function(comm, env) {
  # Assign to global to ensure ordiR2step visibility
  assign("global_comm", comm, envir = .GlobalEnv)
  assign("global_env", env, envir = .GlobalEnv)
  
  mod0 <- rda(global_comm ~ 1, data = global_env)
  mod1 <- rda(global_comm ~ ., data = global_env)
  
  step_res <- ordiR2step(mod0, scope = formula(mod1), 
                         data = global_env, trace = FALSE)
  
  # Cleanup
  rm(global_comm, global_env, envir = .GlobalEnv)
  return(step_res)
}

# --- 3. TEST RUNS ---
data(dune)
data(dune.env)

print("Testing Fail Case...")
tryCatch({
  fail_rda_function(dune, dune.env)
}, error = function(e) print(paste("Caught expected error:", e$message)))

print("Testing Success Case...")
res <- success_rda_function(dune, dune.env)

```

jarioksa:
commit 99346dc referenced wrongly this issue: see instead #787

PR #786, opened 2026-04-20
Bug fix for ordistep scoping issue 

PeetoomHeida:
This seems to fix the issue raised  #785. At the very least it makes the MWE run.

### DIFF
--- a/R/add1.cca.R
+++ b/R/add1.cca.R
@@ -24,13 +24,14 @@
             ## anova.ccalist for 2..scope terms
             if (!is.null(object$CCA)) {
                 nfit <- suppressMessages(
-                    update(object, as.formula(paste(". ~ . +", tt))))
+                    eval(update(object, as.formula(paste(". ~ . +", tt)), evaluate = FALSE),
+                         envir = environment(formula(object))))
                 tmp <- anova(object, nfit, permutations = permutations, ...)
                 adds[i+1,] <- unlist(tmp[2, 5:6])
             }
             else { # first term: simple anova
-                nfit <- update(object,
-                               as.formula(paste(". ~ . +", tt)))
+                nfit <- eval(update(object, as.formula(paste(". ~ . +", tt)), evaluate = FALSE),
+                             envir = environment(formula(object)))
                 tmp <- anova(nfit,  permutations = permutations, ...)
                 adds[i+1,] <- unlist(tmp[1, 3:4])
             }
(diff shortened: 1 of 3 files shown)


jarioksa:
Thanks for this PR and #785. 

It looks indeed necessary to have `update(..., evaluate=FALSE)`.  However, I wonder if it would be better to have `eval.parent(...)` instead of `eval(..., envir = environment(formula(object)))`. Both seem to work in this test case.

jarioksa:
I had a further look at this issue, and it seems that `stats::step()` uses `eval.parent`. Here an extract:

```r
        fit <- update(fit, paste("~ .", change), evaluate = FALSE)
        fit <- eval.parent(fit)
```

This also run embedded in your test code, as did your PR when I changed the code to use `eval.parent()`.

I have struggled with scoping rules for years, and I hope I have already fixed most other issues except this.

Here the changes I tested (and they were OK with your test function). I only tried with `ordiR2step` but I assume `ordistep` would work the same:

```diff
tofu:vegan jarioksa$ git diff
diff --git a/R/add1.cca.R b/R/add1.cca.R
index 0113e589..ff583137 100644
--- a/R/add1.cca.R
+++ b/R/add1.cca.R
@@ -24,8 +24,9 @@
             ## anova.ccalist for 2..scope terms
             if (!is.null(object$CCA)) {
                 nfit <- suppressMessages(
-                    eval(update(object, as.formula(paste(". ~ . +", tt)), evaluate = FALSE),
-                         envir = environment(formula(object))))
+                    eval.parent(update(object,
+                                       as.formula(paste(". ~ . +", tt)),
+                                       evaluate = FALSE)))
                 tmp <- anova(object, nfit, permutations = permutations, ...)
                 adds[i+1,] <- unlist(tmp[2, 5:6])
             }
diff --git a/R/ordiR2step.R b/R/ordiR2step.R
index 0f32ab42..65b3a3c6 100644
--- a/R/ordiR2step.R
+++ b/R/ordiR2step.R
@@ -29,9 +29,10 @@
     ## Get R2 of the scope
     if (R2scope)
         R2.all <- suppressMessages(
-            RsquareAdj(eval(update(object, delete.response(formula(scope)), evaluate = FALSE),
-                            envir = environment(formula(object))),
-                       permutations = R2permutations, ...))
+            RsquareAdj(eval.parent(
+                update(object, delete.response(formula(scope)),
+                       evaluate = FALSE))),
+                       permutations = R2permutations, ...)
     else
         R2.all <- list(adj.r.squared = NA)
     ## Check that the full model can be evaluated
```

PR #789, opened 2026-07-15
summary.isomap: drop reference to undefined 'axes'

sims1253:
summary() on an isomap result aborted because the method evaluated `axes <- min(axes, ncol(object$points))`, although `axes` is neither a formal argument nor defined earlier in the body.

The computed value was never used. Remove the dead assignment so the method can assemble and return the summary object.

Add a regression test that summarizes an isomap object and checks the returned class and component counts.

Found during my https://github.com/sims1253/ry audit

### DIFF
--- a/R/summary.isomap.R
+++ b/R/summary.isomap.R
@@ -1,7 +1,6 @@
 `summary.isomap` <-
 function (object, ...)
 {
-	axes <- min(axes, ncol(object$points))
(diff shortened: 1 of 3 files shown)


jarioksa:
This was indeed overlooked in cf94f090ad41a403c175dce0867e4c95093678f7. I'll fix this.

PR #798, opened 2026-09-14
anova.prc: parse prc call with deparse1 instead of as.character

### LINKED ISSUE #797: Anova on `prc` results will fail if the input variables contained `$` or `(`
Assuming I have a metadata table `md` and a species table `ta`

This works:

```
treatment = md$treatment
time = md$time

mod = prc(ta, treatment, time)
anova(mod, by='onedf')
```

This does not:
```
mod = prc(ta, md$treatment, md$time)
anova(mod, by='onedf')
```
> Error in `str2expression()`:
! <text>:1:7: unexpected '$'
1: ta ~ $
          ^

This also does not (assuming I needed to transpose my species table for vegan)
```
mod = prc(t(ta), treatment, time)
anova(mod, by='onedf')
```
> Error in `as.vector()`:
    ! cannot coerce type 'closure' to vector of type 'any'
    Hide Traceback
    Error in `as.vector()`:
    ! cannot coerce type 'closure' to vector of type 'any'
         ▆
      1. ├─stats::anova(mod, by = "onedf")
      2. ├─vegan:::anova.prc(mod, by = "onedf")
      3. │ └─base::eval(call, parent.frame())
      4. │   └─base::eval(call, parent.frame())
      5. ├─vegan::rda(formula = t ~ treatment * time+ Condition(time))
      6. └─vegan:::rda.formula(formula = t ~ treatment * time + Condition(time))
      7.   └─vegan:::ordiParseFormula(...)
      8.     ├─base::as.matrix(X, rownames.force = TRUE)
      9.     └─base::as.matrix.default(X, rownames.force = TRUE)
     10.       └─base::array(...)
     11.         └─base::as.vector(data)

I suspect both are related to how the formula is built from the arguments provided to `prc`, since the equivalent expression `rda(t(ta)~treatment *time + Condition(time), data=md)` works fine.

gabrielpantoja-cl:
Fixes #797.

## Problem

`anova()` on a `prc` result fails whenever `by` is given and `prc()` was called with anything other than plain symbols:

```r
data(pyrifos)
week <- gl(11, 12, labels = c(-4,-1,0.1,1,2,4,8,12,15,19,24))
dose <- factor(rep(c(0.1,0,0,0.9,0,44,6,0.1,44,0.9,0,6), 11))
md <- data.frame(dose = dose, week = week)

anova(prc(pyrifos, md$dose, md$week), by = "onedf")
#> Error in str2expression(x) : <text>:1:11: unexpected '$'
#> 1: pyrifos ~ $

tp <- t(pyrifos)
anova(prc(t(tp), dose, week), by = "onedf")
#> Error in as.vector(x, mode) : cannot coerce type 'closure' to vector of type 'any'
```

Both cases also emit `Using formula(x) is deprecated when x is a character vector of length > 1`, which is a symptom of the same bug.

## Cause

`anova.prc` recovers the original arguments from `object$call` with `as.character()`. On a language object `as.character()` returns the *components* of the expression rather than the expression as a single string:

```r
as.character(quote(md$dose))  #> "$"  "md"  "dose"
as.character(quote(t(tp)))    #> "t"  "tp"
```

The following `paste()` then vectorizes over that vector and builds a broken formula. In the `t(tp)` case the formula becomes `t ~ ...`, where `t` resolves to the transpose function, hence the `closure` error.

## Fix

Parse the call with `deparse1()` instead. This is the same idiom `prc()` itself already uses for its own arguments (`R/prc.R`), so the two functions are now consistent. `deparse1()` requires R >= 4.0.0 and the package already depends on R >= 4.1.0.

## Verification

Tested against a build of this branch:

| call | before | after |
| --- | --- | --- |
| `prc(pyrifos, dose, week)` | ok | ok |
| `prc(pyrifos, md$dose, md$week)` | error | ok |
| `prc(t(tp), dose, week)` | error | ok |
| `by = "terms"` with `md$...` | error | ok |
| `anova(mod, permutations = ctrl, first = TRUE)` (example in `prc.Rd`) | ok | ok |

The three call styles now give numerically identical results (`all.equal()` over all 45 rows of the `by = "onedf"` table). Row names differ as expected, since the formula terms are named after the expression (`md$dose0.1` instead of `dose0.1`). The spurious `formula(x)` deprecation warning is gone.

`tests/vegan-tests.R` runs without errors.

## Note on tests

I did not add a case to `tests/`, since that would require regenerating `vegan-tests.Rout.save` and the resulting diff would be dominated by R-version noise from my machine rather than by this change. Happy to add one in a separate commit if you prefer.


jarioksa:
Thanks for this. Now merged. Also ported to the release branch (although there is no plan to publish any new 2.7-x release).

